PR #4, opened 2016-02-20
Downgrade Scalaz to 7.1.4.

sellout:
quasar-analytics/quasar runs into issues with 7.1.7, so we downgrade
here to avoid forcing an upgrade there.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,7 +56,7 @@ lazy val standardSettings = Seq(
 
   console <<= console in Test, // console alias test:console
 
-  scalazVersion := "7.1.7",
+  scalazVersion := "7.1.4",
   slcVersion    := "0.4",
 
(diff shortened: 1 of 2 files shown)


sellout:
That should have happened when we downgraded it there, then!

jedesah:
We already do it in quasar for some libraries: https://github.com/quasar-analytics/quasar/blob/master/project/Dependencies.scala#L36

You are right, in retrospect it would have been good to force it with a comment explaining why it was forced. But at the time, no other library was attempting to upgrade it, so just setting back did the trick. In the future I will remember to add force to any library for which I tried to make a patch update that failed for wtv reason.

sellout:
Yeah, cool. I’m trying that in quasar now. I’ll close this if/when that passes. Can you give me an explanation that I can add as a comment?

PR #8, opened 2016-02-25
Adds some tests for `FunctorT`

alissapajer:
namely for `transPrepro`, `transPostpro`, and `transPara`.

I decided to learn about matryoshka, and in doing so, wrote some tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ tmp
 .classpath
 .project
 .settings
+*.swp
+tags
--- a/core/src/main/scala/matryoshka/package.scala
+++ b/core/src/main/scala/matryoshka/package.scala
@@ -293,7 +293,13 @@ package object matryoshka extends CofreeInstances with FreeInstances {
   /** Converts a failable fold into a non-failable, by simply returning the
     * argument upon failure.
     */
-  def once[A](f: A => Option[A]): A => A = expr => f(expr).getOrElse(expr)
+  def orOriginal[A](f: A => Option[A]): A => A = expr => f(expr).getOrElse(expr)
+
+  /** Converts a failable fold into a non-failable, by returning the default
+   * upon failure.
+    */
+  def orDefault[A, B](default: B)(f: A => Option[B]): A => B =
+    expr => f(expr).getOrElse(default)
 
   /** Count the instinces of `form` in the structure.
     */
--- a/CHANGELOG/minor.md
+++ b/CHANGELOG/minor.md
@@ -1,0 +1,2 @@
+* adds unit tests for `transPrepro`, `transPostpro`, and `transPara`
+* renames `once` to `orOriginal` and adds `orDefault`
--- a/core/src/test/scala/matryoshka/spec.scala
+++ b/core/src/test/scala/matryoshka/spec.scala
@@ -123,6 +123,57 @@ object Exp {
   }
 }
 
+sealed trait Exp2[A]
+object Exp2 {
+  case class Const[A]() extends Exp2[A]
+  case class Num2[A](value: Int) extends Exp2[A]
+  case class Single[A](a: A) extends Exp2[A]
+
+  implicit val arbExp2: Arbitrary ~> λ[α => Arbitrary[Exp2[α]]] =
+    new (Arbitrary ~> λ[α => Arbitrary[Exp2[α]]]) {
+      def apply[α](arb: Arbitrary[α]): Arbitrary[Exp2[α]] =
+        Arbitrary(Gen.oneOf(
+          Gen.const(Const[α]),
+          Arbitrary.arbitrary[Int].map(Num2[α](_)),
+          arb.arbitrary.map(Single(_))))
+    }
+
+  def const = Fix[Exp2](Const())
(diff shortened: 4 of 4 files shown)


jdegoes:
:dancers: :tada: Our first third-party contribution to Matryoshka! 

Thanks @alissapajer!

core/src/main/scala/matryoshka/package.scala
```diff
@@ -262,6 +262,12 @@ package object matryoshka extends CofreeInstances with FreeInstances {
     */
   def once[A](f: A => Option[A]): A => A = expr => f(expr).getOrElse(expr)
 
+  /** Converts a failable fold into a non-failable, by returning the default
+   * upon failure.
+    */
+  def onceDefault[A, B](f: A => Option[B], default: B): A => B =
+    expr => f(expr).getOrElse(default)
+
```
sellout: I like this, but it makes me think the `once` name above is wrong. I think this new one could be `orDefault` or `orElse`, while `once` could be renamed `orOriginal` or `orUnmodified`. And is there any reason to make the new one look like `(default: B)(f: A => Option[B])`?

.gitignore
```diff
@@ -10,3 +10,5 @@ tmp
 .classpath
 .project
 .settings
+*.swp
+tags
```
sellout: What’s a swp file?

Tags seems like something that maybe belongs in ~/.config/git/ignore (because it’s relevant to user practice, rather than the specific project).

sellout:
This is great.

One other thing – could you add a file “CHANGELOG/minor.md” with a bulleted list of the changes here? We have a new merge script (created after you submitted this PR) that auto-bumps the semantic version based on the changelog name. I’ll add something about this to the CONTRIBUTING.md file.

.gitignore
alissapajer: vim creates a .swp file when a file is opened for editing in vim. The .swp file saves incremental changes (every 200 characters or 4 idle seconds) so that in the event of a system crash, say, you can recover your unsaved changes to a document open in vim.

I'm happy to put the tags ignoring (and the .swp ignoring) locally in `~/.config/git/ignore` instead of here. They are both user-practice specific, but then so is .DS_Store (Mac users) and .classpath (Eclipse users, I think?). So it really just comes down to where you draw the line.

In any case, I had not heard of the `~/.config/git/ignore` path before, and it's extremely useful! I can definitely move tags there, and if you'd prefer .swp files there too, can do.

core/src/main/scala/matryoshka/package.scala
alissapajer: Re naming, I like `orDefault` and `orOriginal`.

Re function parameters, the signature `(A => Option[B]) => (A => B)` is really the goal here (in both cases), so I think that `(default: B)(f: A => Option[B])` makes that signature more clear.

.gitignore
sellout: You're right – .DSStore, .classpath, .idea (and more?) shouldn't be here either 😄

core/src/main/scala/matryoshka/package.scala
sellout: Great. That all sounds good to me.

.gitignore
garyb: We tend to just use `.*` in the purescript `.gitignore`s, not sure how you feel about that though... the only dotfiles we usually want are `.travis.yml` and `.gitignore` itself.

alissapajer: If we want to go the route: "`.gitignore` should only be for project-level ignores" I think we'd be left with just:
```
project/project
project/target
target
core/target
tmp
```
What I personally usually do is throw everything in .gitignore, including user-specific files (like files from various editors, operating systems, etc) because we know for sure that we don't want to be checking in .swp files or .classpath files, for example. Whereas, it's possible that someone somewhere might want to make a repo where they check in their .swp files, for some reason or other. So it would be awkward for them switching between projects if we force them to put .swp in `~/.config/git/ignore` just so that can have a clean `git status` when working on this project. But maybe that's not a horrible thing to ask someone to do...?

So that's an argument for allowing all files in here that any contributor wants to ignore. We can agree that we don't want any of these files checked in, and so from that point of view, they are "project-specific" because we don't want them in the project :)

jdegoes: > So that's an argument for allowing all files in here that any contributor wants to ignore. 

My 2 cents:

:+1: I agree with this philosophy. It's also the way the front-end repos work. If someone made a change and wanted to stop ignoring certain types of files, because for some reason files with that name / structure became important to the project, we'd want it to be reflected across the whole development team, which wouldn't work if every developer kept such changes in their own private `~/.config/git/ignore`.

sellout: Alright, sounds good – gitignore has everything. I’m fine with that.

sellout:
Sorry, @alissapajer, one last thing – could you squash these commits (you can keep the .gitignore one separate if you want) and rebase on master?

Then I’ll merge it right away. Just have a few hiccups still in our merge process (at least, with 3rd-party contribs), so that’d help.

alissapajer:
@sellout sure thing - rebase and pushed

PR #9, opened 2016-03-04
Add `attributeTopDown`.

sellout:
Stolen from @tpolecat.

### DIFF
--- a/core/src/test/scala/matryoshka/spec.scala
+++ b/core/src/test/scala/matryoshka/spec.scala
@@ -402,6 +402,34 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
       }
     }
 
+    def depth[T[_[_]], F[_]]: (Int, T[F]) => Int = (i, _) => i + 1
+
+    def sequential[T[_[_]], F[_]]: (Int, T[F]) => State[Int, Int] =
+      (_, _) => State.get[Int] <* State.modify[Int](_ + 1)
+
+    "attributeTopDown" should {
+      "increase toward leaves" in {
+        val v = mul(num(0), mul(num(0), num(1)))
+        v.attributeTopDown(0)(depth) must equal(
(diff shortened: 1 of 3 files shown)


core/src/test/scala/matryoshka/spec.scala
```diff
@@ -402,6 +402,34 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
       }
     }
 
+    def depth[T[_[_]], F[_]]: (Int, T[F]) => Int = (i, t) => i + 1
```
mossprescott: Should be `(i, _)`.

core/src/test/scala/matryoshka/spec.scala
```diff
@@ -402,6 +402,34 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
...
+
+    def sequential[T[_[_]], F[_]]: (Int, T[F]) => State[Int, Int] =
+      (x, y) => State[Int, Int](prev => (prev + 1, prev))
```
mossprescott: Similarly, `(_, _)` would work here. Also, it might be cleaner to do something like `State.get <* State.modify(_+1)`, but maybe not when you add the types to make it compile.

mossprescott:
Just some minor stylistic comments in the tests. Even so, these are some great, concise, understandable tests! Nice work.

core/src/test/scala/matryoshka/spec.scala
sellout: Ah, of course. And the ones is `sequential` should be ignored, too. I always end up doing weird stuff when I get confused between non-curried parameters and actual tuples.

sellout:
@mossprescott Updated.

mossprescott:
LGTM

sellout:
Yeah, ok … need to change how the merge script modifies the commit, looks like.

PR #10, opened 2016-03-05
Improve the README a bit.

sellout:
Could use feedback, stuff to copy/paste, etc.

Doing an unconference session later today, so want people to be able to get _some_ direction.


sellout:
Ping @mossprescott (in case there was no other notification that I assigned you)

README.md
```diff
@@ -4,4 +4,63 @@
...
-This library provides generalized folds, unfolds, and traversals for fixpoint data structures in Scala.
+This library provides generalized folds, unfolds, and traversals for fixed point data structures in Scala.
+
+## External Resources
+
+- [Functional Programming with Bananas, Lenses, Envelopes and Barbed Wire](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.41.125) – the iconic paper that collected a lot of this info for the first time
+- [Recursion Schemes: A Field Guide (Redux)](http://comonad.com/reader/2009/recursion-schemes/) – Ed Kmett’s summary (with links to Haskell code) of various folds and unfolds
+- [Unifying Structured Recursion Schemes](http://www.cs.ox.ac.uk/people/jeremy.gibbons/publications/urs.pdf) – a newer paper on how to generalize recursion schemes
+- [Efficient Nanopass Compilers using Cats and Matryoshka](https://github.com/sellout/recursion-scheme-talk/blob/master/nanopass-compiler-talk.org) – Greg Pfeil’s (Scala) talk on this library (and some other things)
+- [Fix Haskell (by eliminating recursion)](https://github.com/sellout/recursion-scheme-talk/blob/master/recursion-scheme-talk.org) – Greg Pfeil’s (Haskell) talk on recursion schemes
```
mossprescott: I guess I would leave out "(Scala)" since it's kind of implied by the mention of "this library", and maybe rephrase as "Greg Pfeil’s talk on recursion schemes in Haskell." For that matter, also "Ed Kmett’s summary of various folds and unfolds, with links to Haskell code." Parentheticals hurt readability.

README.md
```diff
@@ -4,4 +4,63 @@
...
+
+## Introduction
+
+This library is predicated on the idea of rewriting your recursive data structures as functors where the recursive component is replaced by a parameter.
```
mossprescott: Can this be said without using the word "functor"? It's not obvious to me what definition of the term is intended, and you don't yourself define it until the next section.

Also, "the recursive component is replaced by a parameter" is confusing because the _component_ is _always_ a parameter of the (value) constructor, whereas what you mean is something like "the type of the recursive component is given by a type parameter."

README.md
```diff
@@ -4,4 +4,63 @@
...
+
+```scala
+sealed trait Expr
+final case class Var(name: String)          extends Expr
+final case class Lam(v: String, body, Expr) extends Expr
```
mossprescott: Should be `body: Expr` (misplaced comma).

README.md
```diff
@@ -4,4 +4,63 @@
...
+final case class App(f: Expr, arg: Expr)    extends Expr
+```
+could be rewritten as
+
+```scala
+sealed trait Expr[A]
+final case class Var[A](name: String)       extends Expr[A]
+final case class Lam[A](v: String, body, A) extends Expr[A]
```
mossprescott: As above.

README.md
```diff
@@ -4,4 +4,63 @@
...
+final case class App[A](f: A, arg: A)       extends Expr[A]
+```
+
+Then you use one of the fixed point type constructors below to regain your recursive type.
+
+### Fixpoint Types
+
+These types take a one-arg type constructor (a functor) and provide a recursive form of it.
```
mossprescott: I'm still confused about what "functor" (in lower case) means. Wikipedia is no help, so I feel like you might want want to be clear about what you mean:
- really literally any one-arg type constructor?
- something with a `Functor` instance?
- someone's definition you can reference?

mossprescott: For the uninitiated, I think a not-quite trivial example with `Fix` would be a big help here, because it's the one you need to understand first to convert an existing type.

How about defining `ExprF` above and then here showing `type Expr = Fix[ExprF]`, and maybe constructing a value?

README.md
```diff
@@ -4,4 +4,63 @@
...
+- `Fix` – This is the simplest fixpoint type, implemented with general recursion.
+- `Mu` – This is for inductive (finite) recursive structures, models the concept of “data”, aka, the “least fixed point”.
+- `Nu` – This is for coinductive (potentially infinite) recursive structures, models the concept of “codata”, aka, the “greatest fixed point’.
+- `Cofree[?[_], A]` – Does not have a `Corecursive` instance. This represents a structure with some metadata attached to each node. In addition to the usual operations, it can also be folded using an Elgot algebra.
+- `Free[?[_], A]` – Does not have a `Recursive` instance. In addition to the usual operations, it can also be created by unfolding with an Elgot coalgebra.
+
+### Algebras
+
+A structure like this makes it possible to separate recursion from your operations. You can now write transformations that operate on only a single node of your structure at a time.
+
```
mossprescott: Again, a really small example would help tremendously here. Probably a `cata` over some small Expr, but I'm having trouble thinking of one that's very obvious. It would be easier if you had gone with the usual `Expr = Num Integer | Mul Expr Expr` and `eval`

mossprescott:
Sorry I missed this earlier.

This is obviously a big improvement. Other than the syntax errors, feel free to take my comments as suggestions and ignore them if you like.

README.md
sellout: Yeah, I should probably switch to that Expr. I don’t want to use ExprF or make a type alias, because I don’t want encourage users to do that (which can be hard to avoid in a README or tutorial, because you want to construct small explicit examples that never show up in real code). Ideally, there should be one place in their program where they bother to type `Mu[Expr]`, and everything else is generic.

mossprescott: I would have to take your word for it, because in the only real code I've seen `Fix` is all over the place. :laughing:

sellout:
Ok, updated. I _think_ I covered everything. Also added a bit of exposition to the new example. Let me know if it’s any good.

README.md
sellout: Yeah – I need to make a pass over Quasar at some point, and replace all the `Fix` with properly constrained types. Then also s/Fix/Mu/ at the boundaries.

sellout:
Ok, sorry. Just added the Code of Conduct, too. It’s just a small edit to the README, though.

mossprescott:
:+1:

PR #12, opened 2016-03-09
Changes by/for @tpolecat.

sellout:
* Adds `Corecursive[Cofree[?[_], A]]` when `A` has a `Monoid`,
* adds various operations decomposing `{co}elgot`,
* modifies `attributeTopDown` to handle the projection,
* generalizes `ElgotAlgebra`, and
* eliminates `CoelgotAlgebra` (it’s identical to `CoalgebraM`).

### DIFF
--- a/core/src/main/scala/matryoshka/CofreeOps.scala
+++ b/core/src/main/scala/matryoshka/CofreeOps.scala
@@ -17,13 +17,18 @@
 package matryoshka
 
 import scalaz._
-import scalaz.syntax.monad._
 
-sealed class CoalgebraOps[F[_], A](self: Coalgebra[F, A]) {
-  def generalize[M[_]: Monad](implicit F: Functor[F]): GCoalgebra[M, F, A] =
-    self(_).map(_.map(_.point[M]))
+sealed class CofreeOps[F[_], A](self: Cofree[F, A]) {
+  def elgotCata[B](φ: ((A, F[B])) => B)(implicit F: Functor[F]): B =
+    matryoshka.elgotCata(self)(φ)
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/matryoshka/CofreeOps.scala
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package matryoshka
+
+import scalaz._
+
+sealed class CofreeOps[F[_], A](self: Cofree[F, A]) {
```
wemrysi: Is there a difference and/or benefit of `sealed` vs `final` for a concrete class?

sellout: This is the kind of thing where you should just be _telling_ me what to do 😆

I don’t know. I’m sure I copied the exact incantation of how to do implicit conversions from somewhere, but no idea where. Scalaz? In any case, I’m certain someone else was doing it this way, but I don’t know why.

wemrysi: Since you aren't subclassing this type, I'd opt for `final` as it is a bit more precise ('sealed' would allow you to subclass the type in this file), unless we know of some benefit `sealed` provides (I don't know of any, but others might).

wemrysi:
:+1:

core/src/main/scala/matryoshka/CofreeOps.scala
sellout: It looks like Cats uses `final class`, too. So I’m thinking that seems like a good bet.

PR #13, opened 2016-03-12
Class-based algebras.

sellout:
This adds some functionality, but also generalizes things pretty extremely.

I took the approach suggested by @wemrysi and defined a class for the most generic {co}algebras, then used type alias for the more constrained ones. I ended up having to explicitly define a subclass for `GCoalgebra`, because scalac doesn’t like `Id[Id[_]]` showing up in types.

We now

* can `.attribute` any algebra;
* can `.generalize`, `.generalizeElgot`, and `.generalizeM` any {co}algebra (that doesn’t already have a functor in the generalizing position);
* have `transApoM` (for @drostron);
* have totally general {co}algebras (`GElgot*lgebraM`);
* no longer have to define an implicit for each new functor; and
* have some additional docs.

I have a couple issues with this – having to explicitly call an implicit conversion in some cases and _still_ not being able to get `.zip` from implicit conversions.

Also, there are more changes to be made to make the types consistent. But, in general, algebras implicitly convert to/from functions fine, while coalgebras, maybe not so much.

### DIFF
--- a/core/src/main/scala/matryoshka/AlgebraOps.scala
+++ b/core/src/main/scala/matryoshka/AlgebraOps.scala
@@ -16,13 +16,56 @@
 
 package matryoshka
 
-import scalaz._
-import scalaz.syntax.comonad._
+import scalaz._, Scalaz._
 
-sealed class AlgebraOps[F[_], A](self: Algebra[F, A]) {
+final class GAlgebraMOps[W[_], M[_], F[_], A](self: GAlgebraM[W, M, F, A]) {
+  def attribute(implicit W: Comonad[W], M: Functor[M], F: Functor[F]) =
+    matryoshka.attributeGAlgebraM[W, M, F, A](self)
+}
+
+final class ElgotAlgebraMOps[W[_], M[_], F[_], A](self: ElgotAlgebraM[W, M, F, A]) {
+  def attribute(implicit W: Comonad[W], M: Functor[M], F: Functor[F]) =
+    matryoshka.attributeElgotAlgebraM[W, M, F, A](self)
+}
+
+final class GAlgebraOps[W[_], F[_], A](self: GAlgebra[W, F, A]) {
+  def attribute(implicit W: Comonad[W], F: Functor[F]) =
+    matryoshka.attributeGAlgebraM[W, Id, F, A](self)
+
+  def generalizeM[M[_]: Applicative]: GAlgebraM[W, M, F, A] =
+    matryoshka.generalizeM[M, F[W[A]], A](self)
+}
+
+final class ElgotAlgebraOps[W[_], F[_], A](self: ElgotAlgebra[W, F, A]) {
+  def attribute(implicit W: Comonad[W], F: Functor[F]) =
+    matryoshka.attributeElgotAlgebraM[W, Id, F, A](self)
+
+  def generalizeM[M[_]: Applicative]: ElgotAlgebraM[W, M, F, A] =
+    matryoshka.generalizeM[M, W[F[A]], A](self)
+}
+
+final class AlgebraMOps[M[_], F[_], A](self: AlgebraM[M, F, A]) {
+  def attribute(implicit M: Functor[M], F: Functor[F]) =
+    matryoshka.attributeAlgebraM[M, F, A](self)
+
+  def generalize[W[_]: Comonad](implicit F: Functor[F]): GAlgebraM[W, M, F, A] =
+    matryoshka.generalizeAlgebra[W, F, A, M[A]](self)
+
+  def generalizeElgot[W[_]: Comonad]: ElgotAlgebraM[W, M, F, A] =
+    matryoshka.generalizeW[W, F[A], M[A]](self)
+}
+
+final class AlgebraOps[F[_], A](self: Algebra[F, A]) {
+  def attribute(implicit F: Functor[F]) =
+    matryoshka.attributeAlgebra[F, A](self)
+
   def generalize[W[_]: Comonad](implicit F: Functor[F]): GAlgebra[W, F, A] =
-    node => self(node ∘ (_.copoint))
+    matryoshka.generalizeAlgebra[W, F, A, A](self)
 
   def generalizeElgot[W[_]: Comonad]: ElgotAlgebra[W, F, A] =
-    w => self(w.copoint)
+    matryoshka.generalizeW[W, F[A], A](self)
+
+  def generalizeM[M[_]: Applicative](implicit F: Functor[F]):
+      AlgebraM[M, F, A] =
+    matryoshka.generalizeM[M, F[A], A](self)
 }
--- a/core/src/main/scala/matryoshka/AlgebraOps.scala
+++ b/core/src/main/scala/matryoshka/AlgebraOps.scala
@@ -18,54 +18,25 @@ package matryoshka
 
 import scalaz._, Scalaz._
 
-final class GAlgebraMOps[W[_], M[_], F[_], A](self: GAlgebraM[W, M, F, A]) {
-  def attribute(implicit W: Comonad[W], M: Functor[M], F: Functor[F]) =
-    matryoshka.attributeGAlgebraM[W, M, F, A](self)
+final class GElgotAlgebraMOps[E[_], G[_], M[_], F[_], A](
+  self: GElgotAlgebraM[E, G, M, F, A]) {
+
+  def attribute(implicit E: Comonad[E], G: Comonad[G], M: Functor[M], F: Functor[F]) =
+    matryoshka.attribute[E, G, M, F, A](self)
+}
+ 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/matryoshka/algebra.scala
```diff
@@ -0,0 +1,151 @@
...
+
+/** The most general algebra, using both generalized and Elgot comonads as
+  * well as a monad.
+  */
+final class GElgotAlgebraM[E[_], G[_], M[_], F[_], A](f: E[F[G[A]]] => M[A])
+    extends Function1[E[F[G[A]]], M[A]] {
+  def apply(v1: E[F[G[A]]]) = f(v1)
+}
+object GElgotAlgebraM {
+  implicit def zip[E[_]: Functor, G[_]: Functor, M[_]: Applicative, F[_]: Functor]:
```
tpolecat: There is the potential here for `zip` to get shadowed by a local definition since it's such a common identifier, so it might make sense to call this method `gElgotAlgebraMZip` or similar.

sellout: Not intentional as such … but I have never managed to have this implicit found, so I always have to explicitly refer to `GElgotAlgebraM.zip.zip`, with some type parameters thrown in to get it to work. So the short name seemed useful.

I would be happy to give it a long name if it means that implicit search will find it, but no luck so far.

core/src/main/scala/matryoshka/AlgebraOps.scala
```diff
@@ -16,13 +16,27 @@
...
 
-sealed class AlgebraOps[F[_], A](self: Algebra[F, A]) {
-  def generalize[W[_]: Comonad](implicit F: Functor[F]): GAlgebra[W, F, A] =
-    node => self(node ∘ (_.copoint))
+final class GElgotAlgebraMOps[E[_], G[_], M[_], F[_], A](
+  self: GElgotAlgebraM[E, G, M, F, A]) {
 
-  def generalizeElgot[W[_]: Comonad]: ElgotAlgebra[W, F, A] =
-    w => self(w.copoint)
+  def attribute(implicit E: Comonad[E], G: Comonad[G], M: Functor[M], F: Functor[F]) =
```
wemrysi: As `GElgotAlgebraM` is now a class, this could be added as a method therein to avoid having to define this `Ops` class.

core/src/main/scala/matryoshka/AlgebraOps.scala
```diff
@@ -16,13 +16,27 @@
...
+    matryoshka.attribute[E, G, M, F, A](self)
+}
+ 
+final class GAlgebraMOps[G[_], M[_], F[_], A](self: GAlgebraM[G, M, F, A]) {
+  def generalizeElgot[E[_]: Comonad]: GElgotAlgebraM[E, G, M, F, A] =
```
wemrysi: For operations like this that apply to a `GElgotAlgebraM` with particular type arguments, you could still define them as methods and then require implicit evidence that the required type parameters have the necessary value, i.e.
```scala
final class GElgotAlgebraM[E[_], G[_], M[_], F[_], A](f: E[F[G[A]]] => M[A]) {
  def generalizeElgot[EE[_]: Comonad](implicit ev: GElgotAlgebraM[E, G, M, F, A] === GAlgebraM[G, M, F, A]): GElgotAlgebraM[EE, G, M, F, A] =
    new GElgotAlgebraM(matryoshka.generalizeW[E, F[G[A]], M[A]](f))
}
```

core/src/main/scala/matryoshka/CoalgebraOps.scala
```diff
@@ -16,14 +16,20 @@
...
 package matryoshka
 
-import scalaz._
-import scalaz.syntax.monad._
+import scalaz._, Scalaz._
 
-sealed class CoalgebraOps[F[_], A](self: Coalgebra[F, A]) {
-  def generalize[M[_]: Monad](implicit F: Functor[F]): GCoalgebra[M, F, A] =
-    self(_).map(_.map(_.point[M]))
+final class GElgotCoalgebraOps[E[_], G[_], F[_], A](self: GElgotCoalgebra[E, G, F, A]) {
```
wemrysi: Same comments here regarding making these methods on `GElgotCoalgebra`.

core/src/main/scala/matryoshka/algebra.scala
wemrysi: You could also add `zip` as a method on `GElgotAlgebraM` (and then just delegate to it here) which would eliminate the need to explicitly refer to this instance when you aren't working abstractly with some `F[_]: Zip`.

core/src/main/scala/matryoshka/cofree.scala
```diff
@@ -40,6 +40,9 @@ trait CofreeInstances {
   implicit def cofreeShow[F[_], A: Show](implicit F: (Show ~> λ[α => Show[F[α]]])):
       Show[Cofree[F, A]] =
         Show.shows(cof => "(" + cof.head.show + ", " + F(cofreeShow).shows(cof.tail) + ")")
+
+  implicit def toCofreeOps[F[_], A](a: Cofree[F, A]): CofreeOps[F, A] =
```
wemrysi: I think you could make `CofreeOps` an implicit class and avoid this extra def. On that note, you can probably make it a value class as well.

core/src/main/scala/matryoshka/free.scala
```diff
@@ -31,6 +31,9 @@ trait FreeInstances {
           _.point[Free[G, ?]].point[M],
           f(_).map(Free.liftF(_).join))
     }
+
+  implicit def toFreeOps[F[_], A](a: Free[F, A]): FreeOps[F, A] =
```
wemrysi: Same comment as for `toCofreeOps`.

core/src/main/scala/matryoshka/AlgebraOps.scala
sellout: I tried this (fixing the final `E` to be `EE`), but ended up with a complaint that `f` is of type `E[F[G[A]]] => M[A]`, but expected `F[G[A]] => M[A]`.

wemrysi: Ah, sorry, you'll want to use the evidence parameter to convert the value, try
```scala
new GElgotAlgebraM(matryoshka.generalizeW[E, F[G[A]], M[A]](ev(self).f))
```

core/src/main/scala/matryoshka/package.scala
```diff
@@ -312,15 +391,23 @@ package object matryoshka extends CofreeInstances with FreeInstances {
       CorecursiveOps[T, F] =
     new CorecursiveOps[T, F](f)
 
-  implicit def ToAlgebraOps[F[_], A](a: Algebra[F, A]): AlgebraOps[F, A] =
-    new AlgebraOps[F, A](a)
+  implicit sealed class CofreeOps[F[_], A](self: Cofree[F, A]) {
```
wemrysi: Could make this and `FreeOps` `final` classes.

core/src/test/scala/matryoshka/spec.scala
```diff
@@ -466,9 +474,11 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
       }
     }
 
-    def extractFactors: Coalgebra[Exp, Int] = x =>
+    // TODO: Why do we need to explicitly call the implicit here? Is it because
```
wemrysi: Did you happen to try defining `extractFactors: Int => Exp[Int]` and then allowing the implicit to convert at the use site?

sellout: I _think_ I did. Ideally users should be able to use these types or functions interchangeably. I’m not sure if it worked when passed directly to a fold, but I recall scala complaining that `generalize` wasn’t a member of `Int => Exp[Int]`, when trying to convert it to a generalized fold.

wemrysi:
@sellout :+1: (apologies for the delay, I thought I had confirmed this already!).

wemrysi:
@sellout So we're approaching the first anniversary of this PR, not sure if that makes it easier or harder to close ;)

sellout:
Hah, yes – I was about to go over this and the even older PR that’s still open. See if there’s anything worth salvaging – at least open some issues.

paulp:
I would love to see class based algebras, including composable Bialgebras (and Bimonads) as first-class entities.

I think the smart move would be to reduce the function and concept footprint of the current library and create multiple ways to assemble the arguments to a smaller number of functions. A tiny handful of morphisms dominate all uses, and the details of inference and implicits dominate the ergonomics of using the library. If you look at a generalized hylomorphism from different perspectives (input-centric, algebra-centric, monad-centric, adjoint-fold-centric ...) different interfaces emerge. And being explicit about partial application is a tremendous boon for usage because you can lose a lot of the type lambdas and work with single argument type constructors.

PR #16, opened 2016-03-14
Move Arbitrary instances and tests into their own subprojects.

sellout:
This is currently failing to compile, because it claims that “value generalize is not a member of Algebra[Exp,Int]”. I’m not sure why the `ToAlgebraOps` implicit is suddenly not being found. I moved the tests around, but it looks like all the imports and dependencies are in the right spot. If anyone can see the problem, please let me know.


sellout:
Closing, since #37 is merged instead.

PR #17, opened 2016-03-17
Fix broken and add missing prepromorphisms.

sellout:
The behavior of `transPrepro` and `transPostpro` was wrong, only
applying the natural transformation to each node once, when it should
apply them to all deeper nodes before/after each node is processed.

Sorry for any confusion, @alissapajer!

### DIFF
--- a/CHANGELOG/minor.md
+++ b/CHANGELOG/minor.md
@@ -1,0 +1,2 @@
+- add `prepro` and `gprepro`
+- fix signature & behavior of `transPrepro` and `transPostpro`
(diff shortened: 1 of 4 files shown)


sellout:
The previous behavior, applying the natural transformation just once to each node, can be written as
```scala
foo.transCata(addOneƒ ⋘ MinusThree)
```
rather than
```scala
foo.transPrepro(MinusThree, addOneƒ)
```

mossprescott:
¯\_(ツ)_/¯

I can't add much to the comments that have already been made. I'm glad to see some tests, but it isn't very obvious to me what these new/corrected folds are good for. I think it would be great to capture the kind of musing @alissapajer was doing there somewhere as a way to help users wrap their brains around all the variations and what they might be good for, but I don't think scaladoc is the place for that. Maybe the wiki?

sellout:
Yeah, definitely need some more idiomatic examples for these things. I think (among other things) prepro can be used to implement a generalized version of `Semigroup.iterate`. And we’ll also want stuff like that for a performance subproject to benchmark how this approach compares to others.

PR #14, opened 2016-03-14
Add “elgot” versions of various {un}folds.

sellout:
The Elgot generalization of an algebra puts the {co}monad in a different place. E.g.
```scala
F[A] => A
```
normally generalizes to
```scala
F[W[A]] => A
```
but the Elgot generalization is
```scala
W[F[A]] => A
```

PR #18, opened 2016-03-18
Rename `Proj` unapply to `Embed`.

sellout:
Also fix `attributeCoelgotM` to match the correct (generalized) type,
and rename it to `attributeElgotM`.

These are issues I noticed when updating Quasar to use the latest Matryoshka.


sellout:
*sigh* I don’t know how this dropped coverage. I should add a test for `attributeElgotM` anyway, but it seems like it had two uncovered lines before, and still does now.

sellout:
Hrmm, adding a test there is effectively blocked on #14, which fixes the type of the fold that is easiest to test `attributeElgotM` with.

wemrysi:
Looks like you can add the aforementioned test now, :+1: once that's done.

sellout:
I know you gave this a 👍🏾, but wanted to give you a chance to look at the new test …

wemrysi:
@sellout Sorry for the delay, but a belated :+1: anyhow!

PR #19, opened 2016-03-20
We have codata now.

sellout:
`Nu` had basically always been broken. It works now.

Also some minor doc changes.

I would love a better way to write these tests. 1. I either encode the input and output abstractly (`def in`/`def out`) or write it in `Fix` and use `convertTo` to switch to other fixpoints, and 2. they’re all written out explicitly. I would love to have a function that took some `T[F]` and a function like `T[F] => Result` and handled converting the `T[F]` to every fixpoint type and running the function on it. But I don’t know how to write that.


core/src/main/scala/matryoshka/Nu.scala
```diff
@@ -20,25 +20,28 @@ import Recursive.ops._
 
 import scalaz._, Scalaz._
 
-final case class Nu[F[_]](unNu: Id ~> F, a: Id[_])
+/** This is for coinductive (potentially infinite) recursive structures, models
+  * the concept of “codata”, aka, the “greatest fixed point”.
+  */
+sealed abstract class Nu[F[_]]
+private final case class Nu0[F[_], A](unNu: A => F[A], a: A) extends Nu[F]
```
mossprescott: Why is this necessary? I don't see any other constructor for `Nu` here.

sellout: This is how you hide the `A`. I’m open to a better encoding of it, but this is how I figured out[†] how to do GADTs in Scala.

[†]: Read a few lines of a gist.

mossprescott:
This is another one I don't have a ton of insight into.

I'm glad to see that Nu now actually works, and I was inspired to try playing with Mu and Nu (implemented `RenderTree[Mu[F]]`, and watched `convertTo[Nu]` crash).

I agree it would be better to write tests with something like
```scala
sealed trait Runner {
  def run[T[_[_]]: Recursive, F[_]](t: T[F]) => Result
}
def testFix[F[_]](f: Fix[F], r: Runner): Result = 
  r.run(f) and r.run(t.convertTo[Mu]) and r.run(t.convertTo[Nu])
```

But surely you've tried it and it's not that easy ;) For one thing, defining Runners will be ugly.

core/src/main/scala/matryoshka/Nu.scala
wemrysi: You could also use an existential type member
```scala
sealed abstract class Nu[F[_]] {
  type A
  val a: A
  val unNu: A => F[A]
}

object Nu {
  def apply[F[_], B](f: B => F[B], b: B): Nu[F] =
    new Nu[F] {
      type A = B
      val a = b
      val unNu = f
  }
}
```

sellout: @wemrysi Any arguments for or against either? My only argument for the current one is that it is _pretty_ close to `ana = Nu`, which I think is helpful in understanding the structure (and also happens to be how it’s defined in Haskell).

wemrysi: @sellout I'd argue the version with the type member is a more direct way of encoding this in scala (both are almost exactly the same code if you squint, with `Nu.apply == Nu0`).

You could rewrite `Nu.apply` to be `def apply[F[_], B](b: B)(f: B => F[B]): Nu[F]` to make it really look like `ana`.

sellout: My eyes totally glossed over the `object Nu …` part before and I read `new Nu[F] …` as being the definition of `ana`. But yeah, with that `apply`, totally just as good readability-wise. Switching.

sellout:
Updated.

This now uses the “Runner” approach to testing, I didn’t convert everything, but hopefully enough to get a sense of how it works. I can convert the rest if it looks reasonable.

core/src/main/scala/matryoshka/Mu.scala
```diff
@@ -25,23 +25,23 @@ import scalaz._, Scalaz._
...
+  implicit def equal[F[_]: Functor](implicit F: Equal ~> λ[α => Equal[F[α]]]):
       Equal[Mu[F]] =
-    Equal.equal((a, b) => a.convertTo[Fix] ≟ b.convertTo[Fix])
+    Equal.equal((a, b) => F(equal[F]).equal(a.project, b.project))
 
-  implicit def muShow[F[_]: Functor](implicit F: Show ~> λ[α => Show[F[α]]]):
+  implicit def show[F[_]: Functor](implicit F: Show ~> λ[α => Show[F[α]]]):
       Show[Mu[F]] =
-    Show.show(_.convertTo[Fix].show)
+    Recursive.show[Mu, F]
```
mossprescott: Nice!

core/src/test/scala/matryoshka/runners/package.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+  def testCorec[M[_], F[_]: Functor, A](
+    a: A, r: CorecRunner[M, F, A])(
+    implicit Eq0: Equal ~> λ[α => Equal[F[α]]], S0: Show ~> λ[α => Show[F[α]]]):
+      Result =
+    r.run[Fix].apply(a).toResult and
+    r.run[Mu].apply(a).toResult and
+    r.run[Nu].apply(a).toResult and
+    // NB: No Equal for Free, so we can’t test it.
+    // r.run[Free[?[_], Unit]](freeCorecursive[Unit], implicitly, implicitly).apply(a).toResult and
```
mossprescott: I'd say there's no point in showing the implementation if it isn't possible to make it work, especially since it's easy to guess what it would look like from the next line.

core/src/test/scala/matryoshka/runners/package.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+  def testFunc[F[_], G[_]](
+    t: Fix[F], o: Fix[G], r: FuncRunner[F, G])(
+    implicit F: Functor[F], G: Functor[G], Eq0: Equal ~> λ[α => Equal[G[α]]], S0: Show ~> λ[α => Show[G[α]]]):
+      Result =
+    (r.run[Fix].apply(t) must equal(o)).toResult and
+    (r.run[Mu].apply(t.convertTo[Mu]) must equal(o.convertTo[Mu])).toResult and
+    (r.run[Nu].apply(t.convertTo[Nu]) must equal(o.convertTo[Nu])).toResult and
+    // NB: No Equal for Free, so we can’t test it.
+    // (r.run[Free[?[_], Unit]].apply(t.convertTo[Free[?[_], Unit]](F, freeCorecursive[Unit])) must equal(o.convertTo[Free[?[_], Unit]])).toResult and
```
mossprescott: Similar. Worth keeping the comments, though, so nobody wastes time trying.

core/src/test/scala/matryoshka/runners/package.scala
```diff
@@ -0,0 +1,87 @@
...
+    r.run[Nu].apply(a).toResult and
+    // NB: No Equal for Free, so we can’t test it.
+    // r.run[Free[?[_], Unit]](freeCorecursive[Unit], implicitly, implicitly).apply(a).toResult and
+    r.run[Cofree[?[_], Unit]](cofreeCorecursive[Unit], implicitly, implicitly).apply(a).toResult
+
+  abstract class FuncRunner[F[_], G[_]] {
+    def run[T[_[_]]: FunctorT]: T[F] => T[G]
+  }
+  def testFunc[F[_], G[_]](
+    t: Fix[F], o: Fix[G], r: FuncRunner[F, G])(
```
mossprescott: This signature isn't ideal in that it obfuscates what's being tested (by not having `... must_== ...` actually appear in the test), and I believe it will also produce messages with the wrong line number when tests fail (for the same reason).

Any reason this can't be done the same way as the others (i.e. the Runner returns `T[F] => MatchResult[T[G]]`)?

sellout: It can, just requires extra implicits (for `Equal[T[F]]` and `Show[T[F]]`) on the the definition at each test site, so I thought it’d be nice to have a bit less noise in this case. But I don’t mind being consistent with the others, especially with those other points.

core/src/test/scala/matryoshka/spec.scala
```diff
@@ -455,22 +467,28 @@ class FixplateSpecs extends Specification with ScalaCheck with ScalazMatchers {
...
       "evaluate simple expr" in {
-        val v = mul(num(1), mul(num(2), num(3)))
-        v.cata(eval) must equal(6)
-        v.convertTo[Mu].cata(eval) must equal(6)
-        v.convertTo[Nu].cata(eval) must equal(6)
+        testRec(
+          mul(num(1), mul(num(2), num(3))),
+          new RecRunner[Exp, Int] {
+            def run[T[_[_]]: Recursive] = _.cata(eval) must equal(6)
+          })
```
mossprescott: I think this is a win. You don't have to look at `convertTo`, it's more clear that it's intended to work on _all_ Recursive types, and there's a bit less duplication. If we ever add a fourth recursive type, (I don't know, lazily-evaluated Fix? ;), this will make it so we don't have to touch the tests.

A very minor thing is I would expect the types to appear on `testRec` rather than the runner inside it, since they apply to the first argument, too. That would make the Runner's declaration a little less heavy, which would be help readability, I think. Assuming that works, of course.

sellout: Yeah, that doesn’t work, unfortunately. I can do both or just the reader (and DRY, so …).

sellout:
Ok, some more test conversions. I think this is ready for final review.

mossprescott:
:+1:

PR #20, opened 2016-04-06
Upgrade to scalaz 7.2.1

jedesah:
- remove unused imports
- add `Equal` instance for `Cofree`
- remove dependency on `scalaz-specs2` (at least for now) since it is not showing signs of life
- Add `CheckAll` that was provided by `scalaz-specs2`, but could possibly be replaced by a new functions in the `specs2` `3.7.2`
- Add `equal` matcher copied over from `scalaz-specs2`
- Upgrade to version `2.11.8` of Scala
- Find secret recipe for test dependency coherence
- Refactor how dependencies are declared

### DIFF
--- a/core/src/test/scala/matryoshka/spec.scala
+++ b/core/src/test/scala/matryoshka/spec.scala
@@ -18,16 +18,15 @@ package matryoshka
 
 import Recursive.ops._, FunctorT.ops._
 import matryoshka.runners._
+import matryoshka.specs2.scalacheck.CheckAll
 
 import java.lang.String
-import scala.{Boolean, Function, Int, None, Option, Predef, Symbol, Unit},
-  Predef.{implicitly, wrapString}
+import scala.{Boolean, Function, Int, None, Option, Predef, Symbol, Unit}
 import scala.collection.immutable.{List, Map, Nil, ::}
 
 import org.scalacheck._
 import org.specs2.ScalaCheck
 import org.specs2.mutable._
-import org.specs2.scalaz._
 import scalaz._, Scalaz._
 import scalaz.scalacheck.ScalaCheckBinding._
 import scalaz.scalacheck.ScalazProperties._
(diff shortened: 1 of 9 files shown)


jedesah:
I created a `major.md` file because I figured upgrading to scalaz `7.2.1` waranted it, but I didn't realize that we weren't yet on version version `1.0.0`.

sellout:
Yeah, at this point, if we were at 1.0, we’d be at 15.0 😉. I wouldn’t be opposed to the `merge` script being smarter and handling (at least) “major.md” differently for pre-1.0 versions. That way committers don’t have to think about it. Then we could have a “release.md” or something that _only_ works to bump from pre-1.0 to 1.0.

jedesah:
So in the meantime I will rename the file to `minor.md`. I will do that before merging if there is no other comment on the PR to avoid wasting cycles on Travis.

core/src/test/scala/matryoshka/spec.scala
```diff
@@ -175,27 +174,31 @@ object Exp2 {
...
+class Exp2Spec extends Specification with ScalaCheck with CheckAll {
   import Exp2._
 
   implicit val arbExp2Int: Arbitrary[Exp2[Int]] = arbExp2(Arbitrary.arbInt)
   // NB: These are just a sanity check that the data structure created for the
   //     tests is lawful.
-  checkAll(equal.laws[Exp2[Int]])
-  checkAll(functor.laws[Exp2])
+  "Exp2 should satisfy relevant laws" >> {
+    checkAll(equal.laws[Exp2[Int]])
```
sellout: Looks like there should be an `append` here.

core/src/test/scala/matryoshka/spec.scala
```diff
@@ -288,8 +291,8 @@ class MatryoshkaSpecs extends Specification with ScalaCheck with ScalazMatchers
       case _ => t.map(_.head).embed
     }
 
-  "Recursive" should {
-    "isLeaf" should {
+  "Recursive" >> {
+    "isLeaf" >> {
```
sellout: Why the change to `>>` here. I’ve always been a fan of the BDD-style tests, like “X should Y” (despite generally loving operators).

sellout:
Like it. A couple minor comments. And I see you didn’t take my suggestion of `def equals = error` 😉

core/src/test/scala/matryoshka/spec.scala
jedesah: Ah yes, I put one above but not here. Either way works though.

I think I might just remove the one above unless you feel strongly about adding an append here instead.

We seem to use a mix of both styles down below regarding use of `and`.

core/src/test/scala/matryoshka/spec.scala
jedesah: I am not convinced you can nest `should` in `specs2` `3.x`.

I would be inclined to go with this just because they are unit tests and it does not read very well in all cases. But I can try and see if I am imagining things about not being able to nest `should` if you prefer.

jedesah:
@sellout I am not opposed to `def equals = error`. I just figured that it could be made as a separate change since it's not really related anymore now that we reverted back to the `scalaz` `Matcher`. Figured it would be best to think about it a bit more and apply it consistently across the code base.

core/src/test/scala/matryoshka/spec.scala
sellout: Yeah, I’m fine with either. But don’t we eventually have to switch to append/and as we move toward the more functional approach to specs2?

core/src/test/scala/matryoshka/spec.scala
sellout: Ah, well, I wouldn’t worry about it. These shouldn’t be nested anyway – the outer ones should be in their own specifications, but that’s a change for a different PR. At that point, maybe we can change the inners back, and rephrase things so they make more sense (which means better examples in a bunch of places, too).

sellout:
Yeah, I was just kidding about `equals = error`. I think I like it, but it’s a separate change.

core/src/test/scala/matryoshka/spec.scala
jedesah: specs2 made the `Fragments` api more functional by making things `Fragments.foreach` return `Fragments` instead of `Unit` which I think is great. Other things it left as is. I am inclined to squint and call this thing a "dsl" for which mutability was chosen as a technique to implement it. Since I don't like mutability, I would have prefered possibly macros in order to implement the dsl, but I am willing to consider it an implementation detail as it's specs2 responsibility in theory to make sure its mutable dsl works properly.

jedesah:
@sellout Fixed the minor issues. Does this look ready to you?

sellout:
Hey, no fair squashing mid-review 😜

jedesah:
Sorry. They were so minor...

jedesah:
But you're right, should have followed procedure.

sellout:
👍🏾

PR #21, opened 2016-04-17
Generalize to work with directly recursive types.

sellout:
This eliminates operations like `cofCata`, which are now how `cata`
works over `Cofree`.
- add `attbributeColgebra`, `deattribute`, and rename `attribute` to
  `attributeAlgebra`.
- add instances for some `List`, `IList`, `Either`, `\/`, `Option`, and `Maybe`.

Still have issues resolving/unifying instances.


sellout:
I’ve still got _one_ error in the main code:
```
[error] /Users/greg/Documents/SlamData/matryoshka/core/src/main/scala/matryoshka/package.scala:239: value lower is not a member of matryoshka.Recursive[scalaz.Cofree[F,A]]#Base[T]
[error]     fa => Cofree(f(fa ∘ (x => (Recursive[Cofree[F, A]].cata[T](x)(_.lower.embed), x.head))), fa)
```
where it can’t deduce that `Base[T]` is `EnvT[F, A, T]`.

There are also a number of other things I feel like I’ve done in less-than-ideal ways, but I’m not concerned about them if they don’t affect how users use the library (but I won’t be able to see how users are affected until I can get this to compile).

sellout:
I’ve done a lot on this the past couple days. Sorted out most of the difficulties. The one thing left (I hope) is ironing out where we fake mutual recursive stuff (like Diffable and and PotentialFailure). I _think_ I have a path to solving that, too.

sellout:
I fixed the fake mutual recursion (and real mutual recursion is next in line to get sorted out).

There are some other small fixes to make, but mostly there are 50+ errors that look like “could not find implicit value for parameter `T: Corecursive[Fix[Exp]]{type Base[A] = Exp[A]}`”, when it seems like such an instance _does_ exist.

sellout:
So, if I add the following two definitions in a scope that gives the above error (well, with `Mu` instead of `Fix` in the context I tried it), the first one works, and the second says no implicit found:
```scala
val meh = scala.Predef.implicitly[Corecursive.Aux[Mu[Exp], Exp]]
val meh2 = scala.Predef.implicitly[Corecursive[Mu[Exp]] { type Base[A] = Exp[A]}]
```
However, even though the error above uses the second form, it’s always written in the code using the type alias, so it seems like there are some places where the alias is expanded and others where it’s not?

Finally, if I explicitly provide the implicit by passing `scala.Predef.implicitly[Corecursive.Aux[Mu[Exp], Exp]]`, it works. Whereas just using `scala.Predef.implicitly` (so that it infers the `type Base[A]` form) doesn’t work.

So I at least have a workaround now, but it would be nice to not expect users to pass around these implicits everywhere …

sellout:
Ok, so this is now in a working state. Two questions:

1. is it worth keeping `Based` around?
2. can I get rid of having to explicitly supply the implicits for `Corecursive`?

paulp:
It's too big and my understanding is too shallow for me to say much. I could comment better on whether it's keeping Based if you can clarify the comment as indicated. I'd need to perform some of everyone's favorite tedious trial-and-error to develop any insight on losing the explicit implicits and it doesn't look very promising.

paulp:
Implicit parameter lists which have to be supplied explicitly with a peppering of calls to implicitly to obtain the ones which CAN be inferred - to me this is at best a way to discover what the inference problem is, but shipping code like that creates an immediate maintenance and usage nightmare.

sellout:
Updated – I hopefully made the documentation on `Based` clearer. One thing that I didn’t point out before and is perhaps easy to overlook is that the types for `Recursive`, etc. are now shaped like `T` rather than `T[_[_]]`. So you have `Recursive[Fix[F]]` rather than `Recursive[Fix]`. This is why we now have `Base[A]` – because we now need to extract the pattern functor from the proper type.

I also fixed the `Corecursive` resolution problem, so that cleaned up a lot of stuff.

sellout:
Oh, and as for keep/drop `Base`, I’m waffling, because it’s what’s shared between Recursive and Corecursive, _but_ it’s only a type alias declaration, so it’s not like there’s much there.

sellout:
Ok, I’m kind of settled on keeping `Based` now – it’s small, but if we ever implement Birecursive (as I hope to), it’s necessary.

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -184,66 +233,278 @@ import simulacrum._
...
+
+  // NB: The rest of this is what would be generated by simulacrum, except this
+  //     type class is too complicated to take advantage of that.
+
+  def apply[T, F[_]](implicit instance: Aux[T, F]): Aux[T, F] = instance
+
+  trait Ops[T, F[_]] {
+    def typeClassInstance: Aux[T, F]
+    def self: T
+
```
sellout: So, this code is literally copy/pasted from the Simulacrum expansion, with minimum changes to make it work. We can definitely make it cleaner now that it's here (especially these ops that are liable to stick around for a while). But I'd like to make various changes to Simulacrum to make more of it disappear – the first being making it possible to “override” individual generated methods with hand-written ones, which @mpilquist has already said he would like to see happen.

paulp: Ah. Makes sense. Maybe we improve simulacrum then. These days one could delegate the beautification task to [scalafmt](https://github.com/olafurpg/scalafmt).

PR #23, opened 2016-04-29
A few additions around fixed-point type aliases.

sellout:
- Add type aliases for various types (like `Free` and `Cofree`) to show
  the equivalences, and start moving toward fixed-point alternatives.
- Add `AlgebraIso` to make isomorphic algebras explicit.
- A few new pattern functors (in their own package) – `CoEnv`, `ListF`,
  `Diff`, and `PotentialFailue`.
- Add some tests for the Stream alias (mostly just to have a place to put some coalgebras I had written).

### DIFF
--- a/core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
+++ b/core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2014–2016 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package matryoshka.instances
+
+import matryoshka._, Recursive.ops._
+import matryoshka.patterns._
+
+import scala.{Boolean, Int, Option}
+
+import scalaz._, Scalaz._
+
+/** This package provides instances of various common data structures
+  * implemented explicitly as fixed-points.
+  */
+package object fixedpoint {
+  type Free[F[_], A] = Mu[CoEnv[A, F, ?]]
+  type Cofree[F[_], A] = Mu[EnvT[A, F, ?]]
+  type List[A] = Mu[ListF[A, ?]]
+
+  object List {
+    def apply[A](elems: A*) =
+      elems.ana[Mu, ListF[A, ?]](ListF.seqIso[A].reverseGet)
+
+    def fill[A](n: Int)(elem: => A): List[A] =
+      n.ana[Mu, ListF[A, ?]](r => if (r > 0) ConsF(elem, r - 1) else NilF())
+  }
+
+  implicit class ListOps[A](self: Mu[ListF[A, ?]]) {
+    def length: Int = self.cata(size) - 1
+  }
+
+  implicit class RecListFOps[T[_[_]]: Recursive, A](self: T[ListF[A, ?]]) {
+    def headOption: Option[A]              = self.project.headOption
+    def tailOption: Option[T[ListF[A, ?]]] = self.project.tailOption
+  }
+
+  /** A lazy (potentially-infinite) list.
+    */
+  type Colist[A] = Nu[ListF[A, ?]]
+
+  /** A true stream – infinite.
+    */
+  type Stream[A] = Nu[(A, ?)]
+
+  implicit class StreamOps[A](self: Nu[(A, ?)]) {
+    def head: A = self.project._1
+    def tail: Nu[(A, ?)] = self.project._2
+    def drop(n: Int): Nu[(A, ?)] =
+      if (n > 0) tail.drop(n - 1) else self
+    def take(n: Int): Mu[ListF[A, ?]] =
+      (n, self).ana[Mu, ListF[A, ?]] {
+        case (r, stream) if (r > 0) => ConsF(stream.head, (r - 1, stream.tail))
+        case (_, _)                 => NilF()
+      }
+  }
+
+  /** Encodes a function that may diverge.
+    */
+  type Partial[A] = Nu[A \/ ?]
+
+  object Partial {
+    def now[A](a: A): Partial[A] = a.left[Nu[A \/ ?]].embed
+    def later[A](partial: Partial[A]): Partial[A] = partial.right[A].embed
+
+    /** Canonical function that diverges.
+      */
+    def never[A]: Partial[A] = ().ana[Nu, A \/ ?](_.right)
+
+    implicit def monad: Monad[Partial] = new Monad[Partial] {
+      def point[A](a: => A) = now(a)
+      def bind[A, B](fa: Partial[A])(f: A => Partial[B]) =
+        fa.project.fold(f, l => later(bind(l)(f)))
+    }
+
+    /** This instance is not implicit, because it potentially runs forever.
+      */
+    def equal[A: Equal]: Equal[Partial[A]] =
+      Equal.equal((a, b) => (a ≈ b).unsafePerformSync)
+
+    def fromOption[A](opt: Option[A]): Partial[A] = opt.fold(never[A])(now)
+    def fromPartialFunction[A, B](pf: scala.PartialFunction[A, B]):
+        A => Partial[B] =
+      pf.lift ⋙ fromOption
+  }
+
+  implicit class PartialOps[A](self: Nu[A \/ ?]) {
+    import Partial._
+
+    def step: A \/ Partial[A] = self.project
+
+    /** Returns `left` if the result was found within the given number of steps.
+      */
+    def runFor(steps: Int): A \/ Partial[A] =
+      if (steps <= 0) step else step >>= (_.runFor(steps - 1))
+
+    /** Run to completion (if it completes).
+      */
+    def unsafePerformSync: A = self.project.fold(x => x, _.unsafePerformSync)
+
+    /** If two `Partial`s eventually have the same value, then they are
+      * equivalent.
+      */
+    def ≈(that: Partial[A])(implicit A: Equal[A]): Partial[Boolean] =
+      // NB: could be defined as `(self ⊛ that)(_ ≟ _)`
+      (self, that).ana[Nu, Boolean \/ ?](_.bimap(_.project, _.project) match {
+        case (-\/(a1), -\/(a2)) => (a1 ≟ a2).left
+        case (\/-(l1), -\/(a2)) => (l1,      now(a2)).right
+        case (-\/(a1), \/-(l2)) => (now(a1), l2).right
+        case (\/-(l1), \/-(l2)) => (l1,      l2).right
+      })
+  }
+}
--- a/core/src/main/scala/matryoshka/package.scala
+++ b/core/src/main/scala/matryoshka/package.scala
@@ -15,10 +15,12 @@
  */
 
 import matryoshka.Recursive.ops._
+import matryoshka.patterns.EnvT
 
-import scala.{Function, Int, None, Option}
+import scala.{Function, Int, None, Option, Unit}
 import scala.collection.immutable.{List, ::}
 
+import monocle._
 import scalaz._, Scalaz._
 
 /** Generalized folds, unfolds, and refolds. */
@@ -43,7 +45,27 @@ package object matryoshka extends CofreeInstances with FreeInstances {
   type GCoalgebra[N[_], F[_], A] = GCoalgebraM[N, Id, F, A] // A => N[F[A]]
   type CoalgebraM[M[_], F[_], A] = GCoalgebraM[Id, M, F, A] // A => F[M[A]]
   type Coalgebra[F[_], A]        = GCoalgebra[Id, F, A]     // A => F[A]
+  type ElgotCoalgebraM[E[_], M[_], F[_], A] = A => M[E[F[A]]]
+  type ElgotCoalgebra[E[_], F[_], A] = ElgotCoalgebraM[E, Id, F, A] // A => E[F[A]]
 
+  /** An algebra and its dual form an isomorphism.
+    */
+  type AlgebraIso[F[_], A] = Iso[F[A], A]
+  object AlgebraIso {
+    def apply[F[_], A](get: F[A] => A)(reverseGet: A => F[A]): Iso[F[A], A] =
+      Iso(get)(reverseGet)
+  }
+
+  def recCorecIso[T[_[_]]: Recursive: Corecursive, F[_]: Functor] =
+    AlgebraIso[F, T[F]](_.embed)(_.project)
+
+  def lambekIso[T[_[_]]: Corecursive: Recursive, F[_]: Functor] =
+    AlgebraIso[F, T[F]](colambek)(lambek)
+
+  /** There is a fold/unfold isomorphism for any AlgebraIso.
+    */
+  def foldIso[T[_[_]]: Corecursive: Recursive, F[_]: Functor, A](alg: AlgebraIso[F, A]) =
+    Iso[T[F], A](_.cata(alg.get))(_.ana(alg.reverseGet))
 
   /** A NaturalTransformation that sequences two types */
   type DistributiveLaw[F[_], G[_]] = λ[α => F[G[α]]] ~> λ[α => G[F[α]]]
@@ -89,7 +111,7 @@ package object matryoshka extends CofreeInstances with FreeInstances {
   def cofParaM[M[_]] = new CofParaMPartiallyApplied[M]
   final class CofParaMPartiallyApplied[M[_]] { self =>
     def apply[T[_[_]]: Corecursive, S[_]: Traverse, A, B](t: Cofree[S, A])(f: (A, S[(T[S], B)]) => M[B])(implicit M: Monad[M]): M[B] =
-      t.tail.traverseU(cs => self(cs)(f) ∘ ((Recursive[Cofree[?[_], A]].convertTo[S, T](cs), _))) >>= (f(t.head, _))
+      t.tail.traverse(cs => self(cs)(f) ∘ ((Recursive[Cofree[?[_], A]].convertTo[S, T](cs), _))) >>= (f(t.head, _))
   }
 
   /** Composition of an anamorphism and a catamorphism that avoids building the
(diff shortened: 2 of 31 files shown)


sellout:
I kind of forgot to mention the motivation … but we’re starting to use `CoEnv` in Quasar for EJson, and I had just copy/pasted it there for the time being.

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  "fib" should {
+    "begin with 1" in {
+      // TODO: implement `take`
+      (0, 1).ana[Nu, IntStream](fib).project._1 must_== 1
+    }
+
+    "lazily generate the correct sequence" in {
+      // TODO: implement `take`
+      (0, 1).ana[Nu, IntStream](fib).project._2.project._2.project._2.project._2.project._2.project._1 must_== 13
```
mossprescott: Hmm. You can't implement `take`, because `Nu[IntStream]` is always infinite, right?

This is a bit nicer, but only a bit:

```scala
  val fibƒ: Coalgebra[(Int, ?), (Int, Int)] = {
    case (a, b) =>
      val c = a + b
      (c, (b, c))
  }

  val fib = (0, 1).ana[Nu, IntStream](fibƒ)

  // NB: should be `head[T[_[_]], A]: T[(A, ?)] => A`, but types don't infer.
  def head[T[_[_]]: Recursive](xs: T[IntStream]): Int =
    xs.project._1

  def tail[T[_[_]]: Recursive](xs: T[IntStream]): T[IntStream] =
    xs.project._2

  // TODO: use a fold?
  def drop[T[_[_]]: Recursive](n: Int)(xs: T[IntStream]): T[IntStream] =
    if (n > 0) drop(n-1)(tail(xs)) else xs

  head(drop(5)(fib)) must_== 13
```

sellout: Actually, yeah, `drop` is really what I want here. But `take` is implementable – that’s how you get a list from a stream. I’ll add these to a Stream object.

core/src/main/scala/matryoshka/patterns/Diff.scala
```diff
@@ -0,0 +1,73 @@
...
+            left.traverse(f) ∘ (LocallyDifferent(_, right))
+          case Inserted(right) => right.traverse(f) ∘ (Inserted(_))
+          case Deleted(left) => left.traverse(f) ∘ (Deleted(_))
+          case Added(right) => G.point(Added(right))
+          case Removed(left) => G.point(Removed(left))
+          case Unknown(left, right) =>
+            left.traverse(f) ∘ (LocallyDifferent(_, right))
+        }
+    }
+}
```
mossprescott: I'm lost. This won't actually construct Diffs, right? I guess I have to map leaves for my type into `Diff` somehow? An example would really help to make this clear.

mossprescott:
This seems interesting but incomplete, and I'm not really clear on what the standard is for this project. I suppose some of these types are here mostly for didactic purposes, as opposed to actually being used somewhere. I still think it's a good idea to have a working example in a test, at a minimum.

Also lots of TODOs here.

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
mossprescott: Sure, if that's what you mean:

```scala
  type IntListF[A] = ListF[Int, A]

  def take[T[_[_]]: Recursive, S[_[_]]: Corecursive](n: Int)(xs: T[IntStream]): S[IntListF] =
    Corecursive[S].embed(
      (if (n > 0) ConsF(head(xs), take(n-1)(tail(xs)))
      else NilF()): IntListF[S[IntListF]])
```

Still not very sophisticated, though.

core/src/main/scala/matryoshka/patterns/Diff.scala
sellout: Yep. I gotta bring `merga`, etc. over from my Quasar branch.

sellout:
Yeah, I wanted to commit what I already had sketched out locally. Now gotta fill in the gaps and fix coverage.

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
sellout: I _think_ drop and take can be implemented as unfolds to Stream and List, respectively, but I haven't tried it yet.

sellout:
Why, Scaladoc, why?!

mossprescott:
This seems to be building now, apart from codecov. If the report's to be believed, Diff is responsible for much of the missing coverage. I think humans would benefit from some more examples/tests of what Diff does, also.

On the other hand, I think Diff is more a glorified example than a part of the core offering. I guess that's the idea of `patterns`, and I guess that's fine.

I will need some more time to experiment and/or a little help to digest this to offer meaningful feedback. Is this holding up anything you're doing for quasar?

sellout:
The parts I _need_for Quasar I’ve copied over there for the time being. I’m going to try to fix the coverage first, but yeah, more examples helps – prior to the SI-2712 fix, the examples were completely unreadable. They still have more noise than they should, but are way better than they were.

Yeah, patterns maybe blends two things now – one is pattern functors for standard data structures, like ListF, etc. which will become more useful with the `Base` PR, but you can see how the fixed-point aliases in instances.fixedpoint use them already. The other thing is novel uses of fixed-point types, like the Diff pattern functor. I figured it was better that it live here than language on a Quasar branch where it’s been sitting forever.

I’m happy to help with any of this. As I said, there is still more noise than there should be. And I’m happy for any of your experiments to be PRs against this branch. The more tests, the better.

sellout:
Ok, this went from -5% to +5%. That’s a win. I don’t know what this “unexpected coverage changes” thing is. Sounds bogus. The page it takes me to is useless. ¯\_(ツ)_/¯

sellout:
Ok, done fooling with this for now. In review!

BTW, part of what got me started on this (aside from needing some of the bits for Quasar) is wanting to get in a better place for documenting/examples. So that’s at least partially what the `instances.fixedpoint` and `patterns` packages are for.

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
```diff
@@ -0,0 +1,74 @@
...
+      (c, (p1, c))
+  }
+
+  /** Generates an infinite stream of the carrier value.
+    */
+  def constantly[A]: Coalgebra[(A, ?), A] = i => (i, i)
+
+  "fib" should {
+    "begin with 1" in {
+      (0, 1).ana[Nu, (Int, ?)](fib).head must equal(1)
```
mossprescott: This is repetitive and what you're calling `fib` is not really a thing. I would prefer something like:

```scala
/** A coalgebra for generating sequences of values, each calculated from the previous two. */
def iter2[A](f: (A, A) => A): Coalgebra[(A, ?), (A, A)] = { 
  case (a1, b2) => 
    val a3 = f(a1, a2)
    (a3, (a2, a3))
}
/** Infinite sequence of Fibonacci numbers (at least until they overflow int32): `1, 2, 3, 5, ...` */
val fib = (0, 1).ana[Nu, (Int, ?)](iter2(_ + _))
```

If you want it to start with `1, 1`, then you'll have to tack another one on afterward, I suppose.

core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -0,0 +1,70 @@
...
+    def head: A = self.project._1
+    def tail: Nu[(A, ?)] = self.project._2
+    def drop(n: Int): Nu[(A, ?)] =
+      if (n > 0) tail.drop(n - 1) else self
+    def take(n: Int): Mu[ListF[A, ?]] =
+      (n, self).ana[Mu, ListF[A, ?]] {
+        case (r, stream) if (r > 0) => ConsF(stream.head, (r - 1, stream.tail))
+        case (_, _)                 => NilF()
+      }
+  }
```
mossprescott: Nice.

core/src/test/scala/matryoshka/patterns/diff.scala
```diff
@@ -0,0 +1,107 @@
...
+      TwoLists(List(Empty[Mu[Example]]().embed, Empty[Mu[Example]]().embed), List(Empty[Mu[Example]]().embed)).embed.paraMerga(
+        TwoLists(
+          List(NonRec[Mu[Example]]("x", 3).embed),
+          List(Empty[Mu[Example]]().embed, Empty[Mu[Example]]().embed, Empty[Mu[Example]]().embed)).embed)(diff) must
+        equal(CorecursiveOps[Mu, Diff[Mu, Example, ?]](Similar(TwoLists(
+          List(CorecursiveOps[Mu, Diff[Mu, Example, ?]](Different(Empty[Mu[Example]]().embed, NonRec[Mu[Example]]("x", 3).embed)).embed, CorecursiveOps[Mu, Diff[Mu, Example, ?]](Removed(Empty[Mu[Example]]().embed)).embed),
+          List(CorecursiveOps[Mu, Diff[Mu, Example, ?]](Same(Empty[Mu[Example]]().embed)).embed, CorecursiveOps[Mu, Diff[Mu, Example, ?]](Added(Empty[Mu[Example]]().embed)).embed, CorecursiveOps[Mu, Diff[Mu, Example, ?]](Added(Empty[Mu[Example]]().embed)).embed)))).embed)
+    }
+  }
+}
```
mossprescott: These tests don't really convince me that this Diff thing is useful (except as a tricky example). Maybe add a test of the `Show` instance, since it might read in a recognizable way?

mossprescott:
Lots of obviously useful stuff here and I am only suggesting improvements for style and comprehension so feel free to decide it's good enough and move on.

In other words, LGTM.

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
sellout: Yeah, I like that approach. I might call `iter2` something like `binarySequence`(?) and move it out of tests.

And yeah, I get why the sequence doesn’t start where I expect it to. I guess I could seed with `(1, 0)`, but it would be nice to include the seed in the final sequence …

Maybe call it `binarySequenceFrom`?

core/src/test/scala/matryoshka/patterns/diff.scala
sellout: Yeah, I agree, it needs some Show examples.

This was my attempt at RenderTree … with some tradeoffs – like not being able to handle multi-sorted ADTs as easily.

sellout:
Ok, I fixed the things you mentioned, but then I also made a couple other changes – `Partial` and `toTree`, so if you could take a quick glance at the new commit, I’d appreciate it.

core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -67,4 +67,18 @@ package object fixedpoint {
...
+  type Partial[A] = Nu[A \/ ?]
+  type ⊥[A] = Partial[A]
+
+  object Partial {
+    def now[A](a: A): ⊥[A] = a.left[Nu[A \/ ?]].embed
+    def later[A](partial: ⊥[A]): ⊥[A] = partial.right[A].embed
+
+    /** Canonical function that diverges.
+      */
+    def never[A]: ⊥[A] = ().ana[Nu, A \/ ?](_.right)
```
mossprescott: Useful.

core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -67,4 +67,18 @@ package object fixedpoint {
         case (_, _)                 => NilF()
       }
   }
+
+  /** Encodes a function that may diverge.
+    */
+  type Partial[A] = Nu[A \/ ?]
+  type ⊥[A] = Partial[A]
```
mossprescott: Should it be obvious to me why you would want this alias?

sellout: That symbol means bottom/divergence.

core/src/test/scala/matryoshka/patterns/diff.scala
```diff
@@ -104,4 +104,46 @@ class DiffSpec extends Specification with DiffArb with specs2.scalaz.Matchers {
...
+              ||
+              |+- <--- Empty()
+              ||
+              |+- ...
+              ||
+              |+- +++> Empty()
+              ||
+              |`- +++> Empty()
+              |""".stripMargin)
+    }
```
mossprescott: Do me a favor and write it like this, would you?

```scala
 val left = TwoLists(
   List(
     NonRec[Mu[Example]]("foo", 3).embed, 
     SemiRec(2, 
       SemiRec(8, 
         NonRec[Mu[Example]]("meh", 6).embed).embed).embed, 
     Empty[Mu[Example]]().embed,
     Empty[Mu[Example]]().embed),
   List(
     Empty[Mu[Example]]().embed)).embed

val right = TwoLists(
    List(
      NonRec[Mu[Example]]("bar", 3).embed, 
      SemiRec(10, 
        SemiRec(8, 
          NonRec[Mu[Example]]("meh", 7).embed).embed).embed, 
      NonRec[Mu[Example]]("x", 3).embed),
    List(
      Empty[Mu[Example]]().embed, 
      Empty[Mu[Example]]().embed, 
      Empty[Mu[Example]]().embed)).embed

Recursive[Mu].cata((left paraMerga right)(diff))(toTree).drawTree must
...
```

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
```diff
@@ -26,32 +25,30 @@ import org.specs2.mutable._
...
+  val fib = (1, 0).ana[Nu, (Int, ?)](binarySequence(_ + _))
 
   /** Generates an infinite stream of the carrier value.
     */
   def constantly[A]: Coalgebra[(A, ?), A] = i => (i, i)
 
   "fib" should {
     "begin with 1" in {
-      (0, 1).ana[Nu, (Int, ?)](fib).head must equal(1)
+      fib.head must equal(1)
```
mossprescott: Much better.

core/src/test/scala/matryoshka/instances/fixedpoint/stream.scala
```diff
@@ -26,32 +25,30 @@ import org.specs2.mutable._
...
+  /** Infinite sequence of Fibonacci numbers (at least until they overflow
+    * int32)
     */
-  // FIXME: Why does this drop the initial 1?
-  val fib: Coalgebra[(Int, ?), (Int, Int)] = {
-    case (p2, p1) =>
-      val c = p2 + p1
-      (c, (p1, c))
-  }
+  val fib = (1, 0).ana[Nu, (Int, ?)](binarySequence(_ + _))
```
mossprescott: Now _that's_ `fib` (and `(1, 0)` is cute, even).

core/src/main/scala/matryoshka/package.scala
```diff
@@ -316,24 +343,86 @@ package object matryoshka extends CofreeInstances with FreeInstances {
...
+      val c = relation(a1, a2)
+      (c, (a2, c))
+  }
+
+  /** Converts a fixed-point structure into a generic Tree.
+    * One use of this is using `.cata(toTree).drawTree` rather than `.show` to
+    * get a pretty-printed tree.
+    */
+  def toTree[F[_]: Functor: Foldable]: Algebra[F, Tree[F[Unit]]] =
+    x => Tree.Node(x.void, x.toStream)
```
mossprescott: I'm not convinced we would actually use this because the `()`s uglify the result with any sensible Show, but it's clever and that's enough.

sellout: We could define a type isomorphic to `Unit`, like `Hole`, where `show` is `_`.

sellout: Actually, could parameterize the value to use.

core/src/main/scala/matryoshka/package.scala
```diff
@@ -316,24 +343,86 @@ package object matryoshka extends CofreeInstances with FreeInstances {
...
-  implicit def ToCorecursiveOps[T[_[_]]: Corecursive, F[_]](f: F[T[F]]):
-      CorecursiveOps[T, F] =
-    new CorecursiveOps[T, F](f)
+  /** Merges a tuple of functors, if possible.
+    */
+  def mergeTuple[T[_[_]]: Recursive, F[_]: Functor: Merge]:
+      CoalgebraM[Option, F, (T[F], T[F])] =
+    p => Merge[F].merge[T[F], T[F]](p._1.project, p._2.project)
+
+  /** Generates an infinite sequence from two seed values.
```
mossprescott: This probably applies to all of your algabras' comments, but I'll say it anyway:

This comment is inaccurate. This function doesn't generate a sequence. It returns a Coalgebra that can you can use with `ana` to generate a sequence. Or you could use it other ways to do other things, which is the whole point, otherwise you would just apply `ana` here and be done with it. Is there a concise way of describing the algebra that is also true?

core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
mossprescott: Yeah, got that. It's mainly `now(a: A): ⊥[A]` that seems perverse to me. I would expect something called `⊥` to always diverge, but maybe I'm just missing the reference.

core/src/main/scala/matryoshka/package.scala
mossprescott: That would be a little less ugly, but I'm still not sure I would use it. I suppose if it saved writing umpteen RenderTree instances, I might.

When is matryoshka getting heterogeneous types? 😉

core/src/main/scala/matryoshka/package.scala
sellout: I haven’t come up with one yet – but I’d be very happy to adopt one!

One of the reasons for adding the {co}algebra type aliases was to make it clear (by making it possible to jump-to-definition) that they should generally be used with some {un}fold, which I felt made it less disingenuous to describe the algebras that way.

core/src/main/scala/matryoshka/package.scala
sellout: I have kind of convinced myself that the `()` (or `_`) is useful, as it shows where the branches plug into the structure.

mossprescott: I got half way there also, so ¯\_(ソ)_/¯

core/src/main/scala/matryoshka/instances/fixedpoint/package.scala
sellout: Ah, it’s the name used in the literature for this type ¯\\\_(ツ)_/¯

jonsterling: Is this type really the bottom type though? I think `Partial` is a good name... But `_|_` typically refers to a type that has no canonical elements (e.g. whose elements always diverge), not to a type whose elements *may* diverge. (But it's possible it's been used this way–I hadn't seen it though)

sellout: Here’s an example of the usage: http://www.cse.chalmers.se/~nad/publications/danielsson-semantics-partiality-monad.pdf. It’s easier to see in LaTeX, where you can distinguish `⊥` (or even `⊥_{A}`) from `A_{⊥}` (the last being the notion I intended). I guess it also makes more sense in a total language, where `A_{⊥}` is the only form of divergence available.

So I guess the `⊥` alias makes more sense for `Nothing` in Scala. Ok, I fold 😉

jonsterling: Oh, I see... Yeah, the subscripted `\bot` is standard notation from domain theory for adjoining a point/bottom to a dcpo, and makes sense. I think the Scala notation might be a bit tricky to understand, since you can't do subscripts (somebody get on that, lol!)... But now I see what you were going for!

sellout:
Ok, I’m being terrible with this PR. But there’s some more Partial stuff along with the elimination of the alias.

core/src/test/scala/matryoshka/patterns/diff.scala
```diff
@@ -116,34 +116,55 @@ class DiffSpec extends Specification with DiffArb with specs2.scalaz.Matchers {
...
+          List(
+            NonRec[Mu[Example]]("bar", 3).embed,
+            SemiRec(10,
+              SemiRec(8,
+                NonRec[Mu[Example]]("meh", 7).embed).embed).embed,
+            NonRec[Mu[Example]]("x", 3).embed),
+          List(
+            Empty[Mu[Example]]().embed,
+            Empty[Mu[Example]]().embed,
+            Empty[Mu[Example]]().embed)).embed
```
mossprescott: Thank you!

core/src/test/scala/matryoshka/instances/fixedpoint/partial.scala
```diff
@@ -0,0 +1,98 @@
...
+//     "have the given value at an arbitrary point" ! prop { (i: Int, d: Int) =>
+//       i.ana[Nu, (Int, ?)](constantly).drop(20000).head must equal(i)
+//     }
+
+//     "have subsequence of the given value" ! prop { (i: Int, t: Int) =>
+//       i.ana[Nu, (Int, ?)](constantly).take(900) must
+//         equal(List.fill(900)(i))
+//     }
+//   }
+// }
```
mossprescott: Dead code.

core/src/main/scala/matryoshka/package.scala
mossprescott: What do you think of "(Co)algebra for generating an infinite sequence ..."? Or leave it. Actually, yeah, probably leave it as is.

mossprescott:
Just delete the dead code and you have my :+1:. Phew!

PR #26, opened 2016-06-06
Scala.js support

edmundnoble:
Turned `core` into a `CrossProject` and moved tests into its `jvm` subproject.


jedesah:
Out of curiosity. tests need to be in the `jvm project` because of some of the test dependencies being unavailable on both platforms or is that just general practice?

edmundnoble:
@jedesah 

1. You are correct on the first count, as specs2 does not support Scala.js. However, even if this project used ScalaTest 3.0.0 or uTest (Scala.js compatible testing frameworks), 
2. The tests have to be run on *some* platform, be it JS or JVM; typically I test shared code and JVM code on the JVM and test JS-specific things from the JS subproject (splitting it up further if there's points in my code which could function differently on each side, like unsafe casts throwing exceptions or JVM-specific regex), because JVM testing support is better in general (specs2 and pre-3.0.0 ScalaTest, as well as the ability to set breakpoints).

sellout:
So, we should have Travis at least build the Scala.js stuff, even if it isn’t going to be tested at this point.

Could you also add a comment to build.sbt re: specs2 not working on Scala.js yet? I think for Travis we’d test on both once that’s an option.

jedesah:
@edmundnoble Thanks for the explanation. If I understand correctly, if `specs2` were to be available to `scala.js` then we could put the tests in "shared" and run them on whichever platform we wanted or even both. Is that correct?

edmundnoble:
@sellout I believe that Travis is already building the Scala.js version, this is the Travis output that leads me to think that:
```
[info] Packaging /home/travis/build/slamdata/matryoshka/core/js/target/scala-2.11/matryoshka-core_sjs0.6_2.11-0.10.2.jar ...
[info] Done packaging.
```

Please let me know if there's something I'm missing, I don't know much about Travis. I will add the comment to build.sbt shortly.

@jedesah I believe that's the case. https://github.com/typelevel/cats/blob/master/build.sbt is a good example of a project that does just that.

sellout:
@edmundnoble Ah, you’re probably right.

What I’d like to do (but isn’t important in this PR, IMO) is to build JVM and JS in separate Travis instances. This’d be more important (build time-wise) once we run JS tests, but for now I imagine it’s not a big impact.

edmundnoble:
I've added the comment to build.sbt.

sellout:
Thanks, @edmundnoble, this looks good. One other thing I forgot to mention: can you add a file `CHANGELOG/feature.md` (it’ll get renamed to an actual version number by our merge script) with contents something like `- added support for Scala.js`, and squash all your commits into one? Then I’ll merge it.

edmundnoble:
Added `CHANGELOG/feature.md`.

PR #27, opened 2016-06-08
Add some new refolds.

sellout:
This adds `ghyloM`, `dyna`, `codyna`, `codynaM` refolds. As well as a
`ganaM` unfold.

`FunctorT.translate` has been removed, as it is identical to
`FunctorT.transAna` (and identical to `.transCata`, since the traversal
order is irrelevant) and attempting to add `TraverseT.translateM` made it clear
that sometimes you do care about whether its bottom-up or
top-down (because the constraints on `F` and `G` differ), so now always
explicitly use `transAna` or `transCata`.

Fixed the type parameter order for `IdOps.ghylo`.

Added a `Nat` type to `matryoshka.fixedpoint`.

Added type aliases for transforms, and a conversion from Transform to
Algebra.

Added optics for algebras and folds.

Made `freeEqual` available outside of tests.


sellout:
*sigh* once again scaladoc breaks the build.

core/jvm/src/test/scala/matryoshka/instances/fixedpoint/partial.scala
```diff
@@ -16,37 +16,95 @@
...
+    "return now immediately" in {
+      Partial.now(12).unsafePerformSync must equal(12)
+    }
+
+    "return a value when it gets to the end" in {
+      Partial.later(Partial.later(Partial.now(3))).unsafePerformSync must
+        equal(3)
+    }
+
+    // TODO: Should work with any Int, but stack overflows on big negatives.
```
wemrysi: In the implementation of `unsafePerformSync`, have you tried making it `@tailrec` by explicitly `match`-ing on the disjunction rather than `fold`-ing? I'm not sure if the `Ops` class gets in the way of scalac's "tco", but might work.

wemrysi:
:+1: LGTM.

A thought for the fixedpoint instances might be to make them value classes instead of aliases so that you can hang typeclass instances and operations off of them.

core/jvm/src/test/scala/matryoshka/specs2/scalacheck/CheckAll.scala
```diff
@@ -17,16 +17,16 @@
 package matryoshka.specs2.scalacheck
 
 import scala.{Any, StringContext}
-import scala.Predef.Set
+import scala.Predef.{Set}
```
wemrysi: Unecessary braces.

sellout:
Yeah, there’s actually a PR (#13) that converts them to classes. I have to rebase that and fix whatever issues were still there.

core/jvm/src/test/scala/matryoshka/instances/fixedpoint/partial.scala
sellout: I think you’re right that it should be `@tailrec`, but this actually overflows on the non-tailrecable `mc91` (but feel free to tell me how I could make it tailrec).

wemrysi: Yeah, not sure how to make `mc91` tailrec, maybe `Trampoline` to avoid overflow? (No idea if that is possible with `Partial` either!)

sellout:
Ok, addressed your comments, and some other minor stuff (as always), if you want to take a look. Otherwise I’ll merge when the build passes.

PR #28, opened 2016-06-12
The beginnings of mutually-recursive support.

sellout:
This gives us multi-sorted recursion-schemes.

It involves adding a bunch of “higher-order” type classes, etc., which
parallel the ones we already use.

Right now the implementation is pretty minimal – cata/ana, cataM/anaM
are available. These _should_ generalize the same way as the mono-sorted
recursion schemes (para, apo, futu, etc.), but there’s just more stuff
to add before that’s available.

There are a couple minimal tests that show that it does _something_,
that apparently gets the right answer.


sellout:
Blocked by scala/scala#5744 for Scala 2.12 (and scala/scala#5749 for Scala 2.11).

PR #31, opened 2016-07-31
Changes extracted from Quasar.

sellout:


core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -655,6 +655,15 @@ package object matryoshka extends CofreeInstances with FreeInstances {
   def toTree[F[_]: Functor: Foldable]: Algebra[F, Tree[F[Unit]]] =
     x => Tree.Node(x.void, x.toStream)
 
+  /** Replaces all instances of `original` in the structure with `replacement`.
+    *
+    * @group algebras
+    */
+  def substitute[T[_[_]], F[_]](original: T[F], replacement: T[F])(implicit T: Equal[T[F]]):
+      T[F] => T[F] \/ T[F] =
+   tf => if (tf ≟ original) replacement.left else tf.right
```
mossprescott: Somehow this doesn't strike me as a primitive. I guess the two halves are something like `T[F] => (T[F] => Option[T[F])` and `toRight` (i.e. `T[F] => (Option[T[F]] => T[F] \/ T[F])`), but I don't think the former is useful on its own so I guess I don't have a suggestion.

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -696,4 +696,12 @@ package object matryoshka extends CofreeInstances with FreeInstances {
 
   implicit def toFreeOps[F[_], A](a: Free[F, A]): FreeOps[F, A] =
     new FreeOps[F, A](a)
+
+  implicit def equalTEqual[T[_[_]], F[_]: Functor](implicit T: EqualT[T], F: Delay[Equal, F]):
+      Equal[T[F]] =
+    T.equalT[F](F)
```
mossprescott: Could've used this yesterday! Now that I know it exists, let me just go write some more quasar code that will need to be cleaned up when this is merged. ;)

mossprescott:
Codecov seems to think there are some new uncovered lines in EnvT, but I couldn't actually load that page. Is it contributing anything at this point?

mossprescott:
LGTM.

core/shared/src/main/scala/matryoshka/package.scala
sellout: Interesting. Isn’t the former akin to `find(a ≟ b)`? Although, I don’t know how the latter works (i.e., the types are too common).

`orig: T[F] => replacement: T[F] => maybeReplacement: Option[T[F]]`

`orig: T[F] => maybeReplacement: Option[T[F]] => repOrOrig: T[F] \/ T[F]`

So the latter is really `toLeft` then, correct?

Hrmm … shouldn’t the first function just be `orig: T[F] => Option[T[F]]` (or `(T[F] => Boolean) => Option[T[F]]`) then the latter is confusing – Because if the option is `Some`, then we want to throw away its value and use the replacement, otherwise we want to use the unmodifed value.

Anyway, yeah, I dunno what the right decomposition of this is.

sellout:
Yeah, I want to make the coverage pass before merging this. EnvT has a Show instance added, at least.

sellout:
I am bad at writing tests.

sellout:
So, that last commit enabled tests that existed, but weren’t being run before. Not all of those tests were passing, it turns out. I commented out a couple of them and still improved overall coverage, so I guess that’s good.

I also extracted `find` from `substitute` (as discussed above), and added tests for that.

sellout:
Hey, no fair making this “ready” before I even make it “in review” 😉

I added more commits. Thought about squashing some, but the failures are all about test coverage, so the tests that exist pass each step of the way.

Feel free to mark it ready (or even just merge it, using the new process) at this point.

PR #33, opened 2016-08-17
Move fixed-point datatypes to their own package.

sellout:
This adds a _little_ more friction in order to encourage the use of type
class constraints rather than hard-coding a particular type.

### DIFF
--- a/CHANGELOG/breaking.md
+++ b/CHANGELOG/breaking.md
@@ -1,0 +1,1 @@
+- Move fixed-point datatypes to their own package.
--- a/core/jvm/src/test/scala/matryoshka/data/fix.scala
+++ b/core/jvm/src/test/scala/matryoshka/data/fix.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package matryoshka
+package matryoshka.data
 
 import matryoshka.exp.Exp
 import matryoshka.helpers._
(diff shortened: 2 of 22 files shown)


mossprescott:
Seems harmless enough. My only question would be about the package name. You already have `matryoshka.instances`, which would seem to be intended to contain things like this, but there's nothing in it at the moment except the sub-package `fixpoint`, which I guess is more like demos and not really meant to be used in anger, although the name doesn't really convey that. I don't see a clear distinction between `data` and `instances`. Why not just put the concrete types in `instances`?

sellout:
So, `instances` may not be a great name, but I think it does need to be distinct from `data`.

`data` is something you’ll hopefully use sparingly, perhaps just in `main`, when you decide exactly what representation you want when applying all the generic algebras and folds.

`instances.fixedpoint` is more an alternative to (parts of) Cats or Scalaz. It’s an _application of_ recursion schemes that could potentially be a separate library. IIRC, the intervening `instances` name was intended for when #21 was sorted out, and there would be `instances.std` as well as `instances.scalaz` and `instances.cats` in appropriate subprojects.

mossprescott:
OK, well I don't think *either* name is great, but names are hard for this stuff, so as long as you've thought it through I'm willing to defer to you and I don't have any other concerns.

sellout:
Well, the `data` name comes from Cats (with, I think, the same intent) – the type classes are in `cats`, and the data types are in `cats.data`. So, I was just trying to maintain the pattern there. In general, I’ve been taking organizational cues from there instead of Scalaz, even though we still only support Scalaz.

PR #36, opened 2016-10-14
Updates to the build.

paulp:
Primary motivations were an updated kind-projector and an
easy-to-use console (i.e. one which didn't give a bunch of
inappropriate options to scalac.)

Also updated simulacrum and made necessary source changes.
Opened an issue at Monocle to publish new artifacts, because
that's the one holding back an across-the-board upgrade.

  https://github.com/julien-truffaut/Monocle/issues/399

And updated almost every plugin.


paulp:
I don't understand why it failed on travis - doesn't happen like that on my machine. I tried rolling back scalaz.

paulp:
Again failed by codecov. I hate that thing.

build.sbt
```diff
@@ -7,55 +7,55 @@ import org.scalajs.sbtplugin.cross.CrossProject
...
 
-lazy val standardSettings = Seq(
+def universalSettings = Seq(
+  scalaVersion := "2.11.8",
+  organization := "com.slamdata",
+  licenses += ("Apache 2", url("http://www.apache.org/licenses/LICENSE-2.0")),
+  libraryDependencies ++= Seq(
+    "com.github.julien-truffaut" %%% "monocle-core" % monocleVersion,
+    "org.scalaz"                 %%% "scalaz-core"  % scalazVersion,
+    "com.github.mpilquist"       %%% "simulacrum"   % "0.9.0"
```
sellout: Can you update this to 0.10.0? That lets us enable a couple more warts (and _hopefully_ also lets us use 0.10.0 in Quasar to do the same).

build.sbt
```diff
@@ -92,6 +85,9 @@ lazy val standardSettings = Seq(
 //   Wart.ExplicitImplicitTypes - see mpilquist/simulacrum#35
 //   Wart.Nothing
 //   Wart.Throw
+//   Wart.NoNeedForMonad - crashes with a match error after dependency version upgrades
+//     scala.MatchError: (_1: A, _2: T[F])(A, T[F])((a @ _), (tf @ _)) (of class scala.reflect.internal.Trees$Apply)
+//       at org.wartremover.warts.NoNeedForMonad$$anonfun$1$$anonfun$apply$1.apply(NoNeedForMonad.scala:18)
```
sellout: Not necessarily in this PR, but we should be able to switch to `Warts.allBut` now (as we did in Quasar), which means we track the disabled ones explicitly, rather than in comments.

sellout:
I know you have some other local changes for this, so I’ll take another look once that’s in. But I just noticed that this is (apparently) what’s keeping us from simulacrum 0.10.0 in Quasar, so I figured I should review it 😉

paulp:
It's not there yet but I pushed it so you can try it out. I don't really grok the `Cogen` business in scalacheck 1.13 yet. In particular I'm not sure how to deal with the fact that `Arbitrary[A => A]` is not being found anymore and the method which produces them wants a `Cogen[A]` to do it.

You can run it on 2.12 if you want but for the moment you will need to publish a monocle snapshot, since there's no published version against 2.12.

build.sbt
paulp: Yeah that version and all the versions are dialed to the max.

build.sbt
```diff
@@ -7,55 +7,61 @@ import org.scalajs.sbtplugin.cross.CrossProject
...
-  // `scalaz-scalack-binding` is built with `scalacheck` 1.12.5 so we are stuck
-  // with that version
-  "org.scalacheck" %% "scalacheck"                % "1.12.5"       % "test" force()
-)
+def monocleVersion(sv: String) = sv match {
+  case "2.11" => "1.3.0"
+  case _      => "1.4.0-SNAPSHOT"
+}
+def versionDeps(sv: String): Seq[ModuleID] = sv match {
+  case "2.11" => Seq(compilerPlugin("com.milessabin" % "si2712fix-plugin" % "1.2.0" cross CrossVersion.full))
```
sellout: What about using the typelevel compiler for 2.11? Too many differences to make it easy to cross-build?

paulp: Yeah, there's no real motivation and it definitely would complicate cross building. I'd rather wait until they're publishing a 2.12 compiler.

xuwei-k:
`"org.scalaz" %% "scalaz-scalacheck-binding" % "7.2.7-scalacheck-1.13"` released

- https://github.com/scalaz/scalaz/pull/1282
- https://github.com/scalaz/scalaz/wiki/7.2.7

sellout:
It’d be nice to have Travis tests on both Scala 2.11 and 2.12, but since 2.12 only works on JDK8, would we lose the JDK7 tests for 2.11, or is there some way to hand-code the combinations we want? Or maybe Travis just won’t try 2.12 on JDK7, since it knows it doesn’t work?

paulp:
I use this slug in other projects to isolate source based on jvm and scala version - it would work here if necessary.
```scala
  /** Watch out Jonesy! It's the ol' double-cross!
   *  Why, you...
   *
   *  Given a path like src/main/scala we want that to explode into something like the
   *  following, assuming we're currently building with java 1.7 and scala 2.10.
   *
   *    src/main/scala
   *    src/main/scala_2.10
   *    src/main_1.7/scala
   *    src/main_1.7/scala_2.10
   *
   *  Similarly for main/test, 2.10/2.11, 1.7/1.8.
   */
  def doubleCross(config: Configuration) = {
    unmanagedSourceDirectories in config ++= {
      val jappend = Seq("", "_" + javaSpecVersion)
      val sappend = Seq("", "_" + scalaBinaryVersion.value)
      val basis   = (sourceDirectory in config).value
      val parent  = basis.getParentFile
      val name    = basis.getName
      for (j <- jappend ; s <- sappend) yield parent / s"$name$j" / s"scala$s"
    }
  }
```

paulp:
@sellout With that last push I think the Cogen instance is correct, so either the failures are legitimately indicting laws which do not hold or the problem is somewhere else. I'm out for a while.

paulp:
Oops, there is a whole file of other Cogen instances I'd forgotten about. Stupid scalaz glue code muddying my diffs.

tests/src/test/scala/matryoshka/helpers/package.scala
```diff
@@ -27,6 +28,21 @@ import scalaz._, Scalaz._
 import scalaz.scalacheck.ScalaCheckBinding.{GenMonad => _, _}
 
 package object helpers extends SpecificationLike with Discipline {
+  implicit def cogenFree[F[_]: Traverse, A: Monoid : Cogen]: Cogen[Free[F, A]] =
+    Cogen[A] contramap (_.suml)
+
+  implicit def cogenFunctor: Contravariant[Cogen] = new Contravariant[Cogen] {
```
xuwei-k: There is `Contravariant[Cogen]`
https://github.com/scalaz/scalaz/blob/v7.2.7/scalacheck-binding_1_13/src/main/scala/scalaz/scalacheck/ScalaCheckBinding1.scala

dwijnand:
For the TravisCI config you can either:
* add a jdk/scala combination you want: https://github.com/puffnfresh/wartremover/blob/8357e543512dabeb9827a0e8c0675ce89e87c56c/.travis.yml#L7-L10
* or explicitly enumerate the combinations you want: https://github.com/scala/async/blob/ed2a6c160fea527ec83875b813180d6c1fda294d/.travis.yml#L12-L21

paulp:
This'll need a new look.

PR #38, opened 2016-10-24
Update the parts of the build which are easy to update.

paulp:
I've been letting difficult upgrades stand in the way
of easy ones which I really need. Let's stop doing that.
This updates simulacrum to 0.10.0 and kind-projector
to 0.9.2. Everything which wasn't upgraded was left
alone for good reason.

### DIFF
--- a/CHANGELOG/revision.md
+++ b/CHANGELOG/revision.md
@@ -1,0 +1,1 @@
+- Build updates: kind-projector, simulacrum.
--- a/build.sbt
+++ b/build.sbt
@@ -44,7 +44,7 @@ lazy val standardSettings = Seq(
     "JBoss repository" at "https://repository.jboss.org/nexus/content/repositories/",
     "Scalaz Bintray Repo" at "http://dl.bintray.com/scalaz/releases",
     "bintray/non" at "http://dl.bintray.com/non/maven"),
-  addCompilerPlugin("org.spire-math" %% "kind-projector"   % "0.7.1"),
+  addCompilerPlugin("org.spire-math" %% "kind-projector"   % "0.9.2"),
   addCompilerPlugin("org.scalamacros" % "paradise"         % "2.1.0" cross CrossVersion.full),
   addCompilerPlugin("com.milessabin"  % "si2712fix-plugin" % "1.2.0" cross CrossVersion.full),
   ScoverageKeys.coverageHighlighting := true,
--- a/build.sbt
+++ b/build.sbt
@@ -77,7 +77,7 @@ lazy val standardSettings = Seq(
   libraryDependencies ++= Seq(
     "com.github.julien-truffaut" %%% "monocle-core" % monocleVersion % "compile, test",
     "org.scalaz"                 %%% "scalaz-core"  % scalazVersion  % "compile, test",
-    "com.github.mpilquist"       %%% "simulacrum"   % "0.7.0"        % "compile, test"),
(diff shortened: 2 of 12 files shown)


sellout:
Umm 😳 needs a CHANGELOG. Commit with `[skip ci]`, and I’ll merge.

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -35,101 +35,103 @@ import scalaz._, Scalaz._
   * @groupdesc dist Natural transformations required for generalized folds and unfolds.
   */
 package object matryoshka {
+  type Fix[F[_]] = matryoshka.data.Fix[F]
+  val Fix        = matryoshka.data.Fix
```
sellout: I just recently moved `Fix` _out_ of this package, to discourage users from specializing on a fixed-point type. I don’t think it should be brought back in.

The idea being that `Fix` (or `Mu` or `Nu`) should not be referenced in most of the places you use Matryoshka.

paulp: I understand. Have you tried building quasar without this alias? I did.

sellout: We had like one brief day where the `compile` flow in Quasar used `Mu`, but then a change was made that made it difficult to use different fixed-points for different paths. Which is fine, but it meant the other paths being stuck on Fix made the decision of _which_ fixed point for us.

paulp: Ok. I didn't actually mean to leave that alias in there, I'll pull it out. But it is a gargantuan task to adapt quasar, partly because there's no single place to drop a temporary in - many files don't import quasar.Predef - so you have to fix a million files at once.

sellout: Maybe I’ll just finally eradicate Fix from Quasar …

paulp: I must have done those Fix changes in response to something which broke from the other commits, because all I wanted out of this was the plugin updates. Hmmm, don't remember.

paulp: Oh right, it was my local response to trying to test quasar against the plugin-updated matryoshka. Which was completely out of reach in a reasonable amount of time so I came back and added the alias. Didn't mean to push it.

paulp:
Okay I reverted the Fix commit and squashed the last four into one, so it's like I didn't forget the changelog.

core/shared/src/main/scala/matryoshka/package.scala
paulp: TTypes will be your friend.

sellout: @japgolly It’s just that there are type classes for fixed-point types. `Fix` is one specific type, and in general, algebras, folds, etc. can be defined generically – using something like `T[_[_]]: Recursive` (there exists `Recursive[Fix]`, so no worries). Then, you can instantiate it later using `Fix` … or `Mu`, or `Nu`.

So, define everything generically, then specify the type only at the boundaries.

japgolly: @sellout Ah right, so it's more of a concern when defining generic fns (i.e. this lib) but you're saying there's nothing wrong with, say me having types in my own code of `Fix[Blah]`. Is that right?

PR #39, opened 2016-11-14
Add an implicits package.

sellout:
Makes it easier to get all the instances in scope, following Cats’ approach.

This depends on #21.


sellout:
Ok, this is ready for actual review now.

paulp:
Big 👍  from me.

Not-specific-to-PR question: do we perceive some advantage to writing stuff like
```
 package matryoshka.implicits
 
 import matryoshka._
```
as opposed to 
```
package matryoshka
package implicits
```
?

wemrysi:
@paulp Here are [some arguments](https://github.com/quasar-analytics/quasar/pull/1015#r46304405) that were presented against the separate package definitions.

jdegoes:
The main reason is that your brain now has to interpret import statements relative to the entire hierarchy of package declarations, which means it becomes harder to find what you're looking for.

paulp:
If there is an appropriately curated package object at a higher level, it is IMO nicer to get those identifiers all together and at a higher priority than an import. I think there's way too much segmentation of identifiers into places which have to be separately tracked down. There are only two kinds of identifiers:

- ones which you know what they are without checking the origin, because you will sensibly not introduce a second thing with that name in the same project
- the rest

For instance, `import scalaz.concurrent.Task` occurs 99 times in quasar. But there isn't any other Task and it would be foolish to introduce one.

Buuuut, I don't mean to argue about this here or at this time, I just wanted to verify it was done with intent. I'll emulate it.

alissapajer:
@sellout LGTM

PR #40, opened 2016-11-14
Add Birecursive type class.

sellout:
This depends on #21 and #39.


sellout:
This is still in progress, but check the last commit (because the others are from different PRs) and see what you think about this. It’s not too bad – just need to prioritize the instances (and, you know, effectively duplicate them in the first place).

sellout:
Uhh, I guess nevermind with this one – I never changed any of the constraints to `Birecursive`, and of course, once I did, it didn’t work.

PR #41, opened 2016-11-17
Many small changes and cleanup.

sellout:
- a bunch of stuff pulled from quasar.contrib.matryoksha
- renamed `FunctorT#map` and `TraverseT#traverse` to `mapT` and
  `traverseT`, both to match `embedT`/`projectT` and to avoid
  conflicting with Functor/Traverse instances on those types.
- switched to the Typelevel Scala fork and upgraded kind-projector
- cleaned up some constraints, etc. (eg, switching from `lift` to `∘`)
- added some docs


sellout:
I think increasing coverage (while writing docs) has to happen soon …

core/shared/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -77,35 +83,39 @@ package object fixedpoint {
...
 
-    def fillƒ[A](elem: => A): Option ~> ListF[A, ?] =
-      new (Option ~> ListF[A, ?]) {
-        def apply[β](opt: Option[β]) = opt match {
-          case None    => NilF()
-          case Some(b) => ConsF(elem, b)
-        }
-      }
+    // TODO: Needs a better name. Something that evokes `A => B => (A, B)`,
+    //       maybe like `tuple` or `uncurry`?
```
alissapajer: I like `tuple`. I think `uncurry` would look more like `A => B => C => (A, B) => C`?

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -688,6 +714,24 @@ package object matryoshka {
...
+  def partition[A: Order]: Coalgebra[λ[α => Option[(A, (α, α))]], List[A]] = {
+    case Nil    => None
+    case h :: t => (h, (t.filter(_ <= h), t.filter(_ > h))).some
+  }
+
+  def join[A]: Algebra[λ[α => Option[(A, (α, α))]], List[A]] =
+    _.fold[List[A]](Nil) { case (a, (x, y)) => x ++ (a :: y) }
+
+  // NB: not in-place
+  def quicksort[A: Order]: List[A] => List[A] = {
```
alissapajer: I didn't see any tests for `quicksort`, `join`, and `partition`. If we're supporting them in the library it would be good to test them, maybe with scalacheck?

core/shared/src/main/scala/matryoshka/patterns/ListF.scala
```diff
@@ -52,6 +52,14 @@ object ListF {
     case NilF()      => Seq.empty
   } (s => s.headOption.fold[ListF[A, Seq[A]]](NilF())(ConsF(_, s.tail)))
 
+  def takeUpTo[N, T, A](implicit N: Recursive.Aux[N, Option], T: Recursive.Aux[T, ListF[A, ?]]): Coalgebra[ListF[A, ?], (N, T)] =
+    pair => pair._1.project.fold[ListF[A, (N, T)]](NilF())(p => pair._2.project.map((p, _)))
+
+  def find[A](cond: A => Boolean): Algebra[ListF[A, ?], Option[A]] = {
```
alissapajer: I think we could generalize this away from `ListF`, to a more basic notion of `find` for some recursive structure. Not for this PR, but I think that sort of function, might be a way to help users more easily find the correct morphism.

I wonder if that should be a second library? Or at least a different package here. Basically, base matryoshka contains the impls of the morphisms, and then the usages package wraps the direct calls to the morphisms into functions with more commonly used signatures.

alissapajer: There are common patterns we see in the connectors for example, with `interpretM` and `freeCata`. 

A good approach could be to ask the question: "What questions do people answer with matryoshka?" For example: "I'd like to find all instances of a certain pattern and accumulate those (mapped to another type perhaps) into a list. How do I do that?" And then we wrap that all up in a function that calls `para` that takes as input the pattern to match on.

core/shared/src/main/scala/matryoshka/patterns/CoEnv.scala
```diff
@@ -22,9 +22,17 @@ import scalaz._, Scalaz._
 
 /** The pattern functor for Free. */
 final case class CoEnv[E, F[_], A](run: E \/ F[A])
+
 object CoEnv extends CoEnvInstances {
   def coEnv[E, F[_], A](v: E \/ F[A]): CoEnv[E, F, A] = CoEnv(v)
 
+  def hmap[F[_], G[_], A](f: F ~> G) =
```
sellout: “higher-kinded”, as in https://github.com/pa-ba/compdata/blob/master/src/Data/Comp/Multi/HFunctor.hs#L89 (these are just moved from ones we wrote in Quasar).

More of these type classes exist in #28, but the Cats naming scheme might lead us to `FunctorK`, etc. instead. Finally, there is work on adding polykinds to scalac, which would mean that `FunctorK` and `Functor` can be unified into a single type class, and then this operation becomes just `map` … which is actually a bit of a problem, because we have both `HFunctor` _and_ `Functor` on this type currently …

core/shared/src/main/scala/matryoshka/patterns/ListF.scala
sellout: I’m generally pretty wary about wrapping things up with their folds. I do it in `matryoshka.instances.fixedpoint` to illustrate how we can implement existing interfaces, but one of the big benefits of recursion schemes is the compositionality, and hiding the algebras (as we used to do in Quasar, but have moved away from) removes that. That’s why (for the most part) the algebras used _by_ `matryoshka.instances.fixedpoint` are defined in `matryoshka.patterns` – you generally want to try to work with patterns instead of instances.

sellout: Oh, and BTW, as of a couple PRs ago, `freeCata` is dead – it’s now just `cata`.

sellout:
Ok, updated this. Adding a test for quicksort was helpful, because it revealed an awkwardness in defining it as a function rather than a method (would have had to call `quicksort[Int].apply(someList)`).

sellout:
This could use another review – want to try to update Quasar after this gets merged (just got frustrated by not being able to use `Free#para` 😉)

alissapajer:
LGTM - merging!

PR #47, opened 2016-12-06
Fix example in README.md

vil1:
The example in README.md wasn't compiling against the latest published version (0.15.1).

The necessary functor could be freer but honestly, I don't know how to use Coyoneda to grab a free functor.

I switched to `CorecursiveT` because it aligns with the `T[_[_]]` needed here, whereas `Corecursive`doesn't.

This was an attempt for me to understand what's going on here, but it wasn't completely successful. So please disregard this if it looks stupid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,7 @@ This diagram covers the major classes of transformations. The most basic ones ar
 Here is a very simple example of an algebra (`eval`) and how to apply it to a recursive structure.
 
 ```scala
-
-// we will need a Functor[Expr] in order to call embedT bellow
+// we will need a Functor[Expr] in order to call embed bellow
 implicit val exprFunctor = new scalaz.Functor[Expr] {
   override def map[A, B](fa: Expr[A])(f: (A) => B) = fa match{
     case Num(value) => Num[B](value)
--- a/README.md
+++ b/README.md
@@ -73,14 +73,25 @@ This diagram covers the major classes of transformations. The most basic ones ar
 Here is a very simple example of an algebra (`eval`) and how to apply it to a recursive structure.
 
 ```scala
+
+// we will need a Functor[Expr] in order to call embedT bellow
+implicit val exprFunctor = new scalaz.Functor[Expr] {
+  override def map[A, B](fa: Expr[A])(f: (A) => B) = fa match{
+    case Num(value) => Num[B](value)
+    case Mul(l, r) => Mul(f(l), f(r))
+  }
+}
+
 val eval: Algebra[Expr, Long] = { // i.e. Expr[Long] => Long
   case Num(x)    => x
   case Mul(x, y) => x * y
 }
+ 
+def someExpr[T[_[_]]: CorecursiveT]: T[Expr] =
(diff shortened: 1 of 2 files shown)


sellout:
Whoops, thanks. Really have to get this stuff using tut …

vil1:
You're completely right, and that's sounds like a thing I could work on !
Maybe not this week-end (birthday party incoming), but the next one.
 Expect some more PR's

README.md
```diff
@@ -73,14 +73,25 @@ This diagram covers the major classes of transformations. The most basic ones ar
...
+}
+
 val eval: Algebra[Expr, Long] = { // i.e. Expr[Long] => Long
   case Num(x)    => x
   case Mul(x, y) => x * y
 }
+ 
+def someExpr[T[_[_]]: CorecursiveT]: T[Expr] =
+  Mul(Num[T[Expr]](2).embedT, Mul(Num[T[Expr]](3).embedT,
+      Num[T[Expr]](4).embedT).embedT).embedT
```
sellout: Glad you were able to figure out how to fix it. This definitely works, although a more general version is
```scala
def someExpr[T](implicit T: Corecursive.Aux[T, Expr]): T =
  Mul(Num[T](2).embed, Mul(Num[T](3).embed,
      Num[T](4).embed).embed).embed

someExpr[Mu[Expr]].cata(eval)
```
The `T[_[_]]` type is more restrictive – basically only allowing Fix/Mu/Nu, because it needs a type that can be written with a functor-shaped hole. Whereas `T` (with the `Corecursive.Aux[T, Expr]` implicit) is any proper type from which you can _extract_ an Expr functor – that clearly includes `Fix[Expr]`, but could also include, say, a directly-recursive `ExprR` that had a mapping to the `Expr` functor (note that `ExprR` is a proper type that doesn’t have a type parameter for `Expr` anywhere, so the `T[_[_]]` couldn’t match it).

Also, with an upcoming PR, even the `T[_[_]]` version would allow you to just write `.embed` rather than `.embedT`.

I eventually want to get rid of the `CorecursiveT` and other `*T` type classes, but at least until we have mutual recursion they are still useful.

vil1: I've seen the comments above `CorecursiveT` so I suspected it was a sub-optimal solution.
I'll do my best making sense out of all the words you've written there (it's still a bot hard for me), but meanwhile I'll test teh solution you propose on my local project (manual tut eh) and modify the PR.

vil1:
Well, that works fine.
And for someone like me who's accustomed with the Aux pattern, that seems to make a little bit more sense (still very blurry though, but I'll keep trying)

README.md
```diff
@@ -73,16 +73,27 @@ This diagram covers the major classes of transformations. The most basic ones ar
 Here is a very simple example of an algebra (`eval`) and how to apply it to a recursive structure.
 
 ```scala
+
+// we will need a Functor[Expr] in order to call embedT bellow
```
sellout: This should say `embed` instead of `embedT` now (and there’s an extra blank line above it)

vil1:
Here you go !

PR #49, opened 2016-12-08
microsite skeleton

tpolecat:
This sets up tut with the [sbt-microsites](https://47deg.github.io/sbt-microsites/) plugin which not only compiles the tut doc but puts it into a little website that looks nice. You can push it straight to a gh-pages branch and it will show up at `https://slamdata.github.io/matryoshka/`.

To make it work you need to install Jekyll and create a gh-pages branch, both of which are documented at the link above. You can then

    sbt makeMicrosite

and then

    cd docs/target/site
    jekyll serve

and preview the site at http://localhost:4000/matryoshka/

If you want a few examples I did the [atto](https://tpolecat.github.io/atto/) and [Tuco](https://tpolecat.github.io/tuco/) sites this way.


tpolecat:
```
[info] Fix should
[info]   satisfy relevant laws
[info]     equal must satisfy
[info]       + equal.commutativity
[info]       + equal.reflexive
[error]       x equal.transitive
[error]        Gave up after only 17 passed tests. 87 tests were discarded. (fix.scala:31)
[info] 
```

sellout:
@tpolecat Could you add a file named "CHANGELOG/revision.md" containing something like "- added skeleton for tut-based docs"?

tpolecat:
As you wish.

Issue #45, opened 2016-11-30
“Adjust” `FunctorT` for directly-recursive types.


sellout:
`FunctorT` (and `TraverseT`) was originally created to support transformations on types that had some way to “map” over the pattern functor, but lacked either a `project` or `embed` (i.e., {Co}Free). However, there’s no way to change those type classes to the directly recursive approach and maintain that property. I’m not sure what the right approach is going forward.

1. {Co}Free no longer works with `FunctorT`, so the original use case is already gone (the old implementations relied on incorrect semantics for those types, so it’s just as well)
2. But _maybe_ there are other types that would benefit from that approach?
3. With some constraint changes, we could fold them into the `Recursive` and `Corecursive` type classes.
4. But at that point, there are just generic algebra transformations that would avoid needing additional folds at all.

Ok, so I think I’ve convinced myself that those type classes should go away and we should just implement algebra transformations (e.g., `AlgebraicTransform => Algebra`) to allow us to continue to write transformations the way we already do.

paulp:
> But maybe there are other types that would benefit from that approach?

I won't pretend I understand all the dimensions of the problem here, but I understand software well enough to say definitively that nothing good has ever come of allowing an italicized maybe to affect one's design.

sellout:
Yeah, that’s what the conclusion says, too 😉

paulp:
Right - but *maybe* my generalized version will find more applications elsewhere. 😃

sellout:
To be clear (because, upon rereading I think it’s not) my plan is to eliminate the old approach that allows for types that can implement some `mapT` but lack either a `project` or `embed` and to also eliminate the `trans*` operations, replacing them with a smaller number of algebra transformations (which, actually, already exist in some cases because they were useful with the old approach), so we can still write `F[T[G]] => G[T[G]]` style transformations, but easily use them in a context that expects an `F[T[G]] => T[G]` style fold.

sellout:
With some (not yet PRed) changes that redefine all of the `trans*` operations in terms of the {co}algebraic operations, these become even more obviously `transform <<< project` and `transform >>> embed`. If inference were better, that’d be enough for me. But … we usually have to jump through some hoops to get `<<< project` to actually work (like, not using point-free, adding some type annotations, etc.), so a few helpers like `Transform => Algebra` are probably still useful.

Issue #44, opened 2016-11-30
Add `Birecursive` type class


sellout:
The direct approach of
```scala
sealed trait Birecursive[T] extends Recursive[T] with Corecursive[T]
```
doesn’t seem to cut it when an abstract type member is in play, so we need something more clever.

I, @paulp, and @edmundnoble have taken swipes at this so far.

sellout:
There is [a gist](https://gist.github.com/sellout/3bec3b9d3b6de2fd224daee5cf859ed6) with a minimal version of the problem.

PR #51, opened 2016-12-11
Eliminate FunctorT and TraverseT type classes.

sellout:
As well as other changes needed in the wake of the directly-recursive
overhaul – like fixing `EqualT`/`ShowT` implicits, etc.


sellout:
@paulp This is “in progress” because there _may_ still be a couple changes as I finish updating Quasar to use it, but it should be pretty solid.

(And apologies to @tpolecat for breaking `EqualT`, etc. for a bit, there)

paulp:
Looks like a great simplification so far.

sellout:
Yeah, so some comments in that vein …

1. `RecursiveT`/`Co…T`/`Bi…T` still exist for a few reasons – they make it easy to slowly transition Quasar, they are helpful in faking mutual recursion until that is in, and `BirecursiveT` works while `Birecursive` doesn’t, so that’s also nice;
2. having `Birecursive` working (#44) would help clean up a bunch of stuff; and
3. @milessabin’s work on multiple implicit lists will clean up _soooo_ much stuff.

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -301,6 +256,185 @@ trait Recursive[T] extends Based[T] {
     (implicit R: Corecursive.Aux[R, Base], BF: Functor[Base])
       : R =
     cata[R](t)(R.embed(_))
+
+  def mapR[U, G[_]: Functor]
```
sellout: This and below is pulled from `FunctorT` and `TraverseT` – it seemed prudent to not get rid of the operations _yet_ (#45 is tracking this), since we use them a bunch in Quasar and I’m just not sure what the right approach is.

There are also some operations moved from earlier in the type class – into the group of things that should be in `Birecursive` once that is actually usable.

core/shared/src/main/scala/matryoshka/implicits/IdOps.scala
```diff
@@ -28,9 +28,9 @@ sealed class IdOps[A](self: A) {
     matryoshka.hyloM(self)(f, g)
 
   object ghylo {
-    def apply[W[_], N[_]] = new Aux[W, N]
+    def apply[W[_], N[_]] = new PartiallyApplied[W, N]
```
sellout: Renamed these “helper” classes to match the Cats approach, which is named this way to avoid confusion with the _other_ `Aux` pattern. Figured since I was adding new ones, I should make sure the code is consistent.

This is something that multiple implicit lists would make disappear.

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -94,58 +94,52 @@ package object matryoshka {
...
   /** Transform a structure `F` containing values in `W`, to a structure `G`, 
     * in bottom-up fashion.
     * @group algebras 
     */
-  type GAlgebraicTransform[T[_[_]], W[_], F[_], G[_]]        = F[W[T[G]]] => G[T[G]] // GAlgebraicTransformM[T, W, Id, F, G]
-  /** Transform a structure `F` to a structure `G`, in bottom-up fashion.
-    * @group algebras 
-    */
-  type AlgebraicTransform[T[_[_]], F[_], G[_]]               = F[T[G]] => G[T[G]]    // GAlgebraicTransformM[T, Id, Id, F, G]
+  type AlgebraicGTransform[W[_], T, F[_], G[_]]        = F[W[T]] => G[T]
```
sellout: These new aliases just use the new proper type approach. It means the differ a bit (e.g., the types of `AlgebraicTransform` and `CoalgebraicTransform` are now the same, so they’re just `Transform`).

tests/src/test/scala/matryoshka/instances/fixedpoint/partial.scala
```diff
@@ -69,7 +69,7 @@ class PartialSpec extends Specification with ScalazMatchers with ScalaCheck with
     "return after multiple runs" >> prop { (a: Conat, b: Conat) =>
       val (ai, bi) = (a.cata(height), b.cata(height))
       bi > 0 ==> {
-        val first = (a + b).transAna(Partial.delay(27)).runFor(ai)
+        val first = (a + b).transAna[Partial[Int]](Partial.delay(27)).runFor(ai)
```
sellout: These operations (like the non-`trans` operations with the same names) now require a type annotation, because they only know that the result type has a particular functor, but not what proper type that functor has been extracted from.

sellout:
Ok, @paulp _now_ it‘s yours.

Once this gets in, I’ll submit the PR that updates Quasar to use this version.

paulp:
@sellout shouldn't `CorecursiveT` also lose its `@typeclass` annotation?

paulp:
@sellout what about 06e82db4dc, is that okay?

sellout:
Ah yes, that is great :)

paulp:
I think that's everything concrete I have words for at the moment. Do you want me to make the CorecursiveT change?

sellout:
Sure. I like this new thing in the space between code review and pairing 😄

paulp:
@sellout I think it's called code repairing

paulp:
I'll merge it after the tests pass.

PR #52, opened 2016-12-12
Delay type class.

paulp:


paulp:
I'm just offering this because I have it and type aliases make me vaguely uncomfortable. I don't feel strongly about its necessity, so please merge or close as you see fit.

core/shared/src/main/scala/matryoshka/Delay.scala
```diff
@@ -0,0 +1,31 @@
...
+
+/** To avoid diverging implicits with fixed-point types, we need to defer the
+  * lookup.
+  */
+trait Delay[F[_], G[_]] {
+  def apply[A](fa: F[A]): F[G[A]]
+}
+
+object Delay {
+  implicit def fromNT[F[_], G[_]](nt: F ~> (F ∘ G)#λ): Delay[F, G] =
```
sellout: I assume this is implicit to avoid having to rewrite all the Delay instances? Maybe a comment that we’d like this to be _explicit_ in the future?

sellout:
Yeah, I like this. Never did like the implicit natural transformation thing (although, we still have that for the time being – but now we have a path away from it).

core/shared/src/main/scala/matryoshka/Delay.scala
paulp: Yeah, I'll add a comment to that effect.

PR #55, opened 2016-12-22
Make RecursiveT serializable

rabbitonweb:
I'm using Matryoshka on Spark. Becasue computation has to be send to
cluster over the network, whole code needs to serialize. Becasue
RecursiveT did not extend Serializable, this was not possible. This
commit changes that.

### DIFF
--- a/core/shared/src/main/scala/matryoshka/RecursiveT.scala
+++ b/core/shared/src/main/scala/matryoshka/RecursiveT.scala
@@ -16,6 +16,7 @@
 
 package matryoshka
 
+import scala.Serializable
 import scalaz._, Scalaz._
 
 /** This is a workaround for a certain use case (e.g.,
@@ -23,7 +24,7 @@ import scalaz._, Scalaz._
   * Define an instance of this rather than [[Recursive]] when possible.
   */
 // NB: Not a `@typeclass` because we don’t want to inject these operations.
-trait RecursiveT[T[_[_]]] {
+trait RecursiveT[T[_[_]]] extends Serializable {
   def projectT[F[_]: Functor](t: T[F]): F[T[F]]
 
   def cataT[F[_]: Functor, A](t: T[F])(f: Algebra[F, A]): A =


sellout:
Have you tried this with publishLocal? I ask, because I figure you’ll probably run into needing `Recursive` and `Based` as well, if not `BirecursiveT`, `CorecursiveT`, and `Corecursive` too.

rabbitonweb:
I've publlished local and run it. Was fine with Matryoshka but failed on scalaz instances. Once those are resolved I will get back to this PR.

But I guess if we are chanignig this typeclass, we should also change others

sellout:
I don’t know much about Serializable, other than that Spark infects everything with it. I haven’t seen `extends Serializable` on a type class before, so I’m a bit hesitant (but, obviously, we need _some_ solution). Is it possible that there’s a data type somewhere that is missing `extends Product with Serializable`, and so it just _looks_ like we need it on the type class?

rabbitonweb:
I guess you have not seen it because u never had to send instance over the network :)
Spark was complaining about instances of this trait. Basically it tried to serialize the method (`cataT` I think) and the machinery behind lambda serialization builds a closure around the instance where the method lives and requires the serialization of that instance as well.

To be honest, as mentioned, I'm having same issues with Scalaz code where I can not add `extends Serializable` - once I know how to handle those I will get back to Matryoshka

sellout:
“I guess you have not seen it because u never had to send instance over the network”
where “u” = “anyone using Scalaz or Cats” (which, admittedly, is a tiny fraction of Scala users) since those libraries haven’t added it to any type classes.

So, you mention `cataT`, which I don’t think is referenced in Quasar code directly, right? Which gives me a hint that the problem here is not necessarily type classes in general, but only our slightly magical type classes … maybe. We have an implicit conversion from `RecursiveT` to `Recursive`, and that should be the only place that `cataT` is ever referenced. So, I have a _feeling_ that it makes `RecursiveT` not _quite_ a type class. I’m doing a lot of hand-waving here, but it makes me think that maybe the `*T` type classes should be `Serializable` but not the others. Which is nice, because those type classes are getting ready for the chopping block anyway 😉

rabbitonweb:
I will send spark stack here, it rather nicely shows the path of what failed in serialization

rabbitonweb:
Duplicated by https://github.com/slamdata/matryoshka/pull/57

PR #56, opened 2017-01-06
Add more learning material

Fristi:
Adds a excellent talk about recursion schemes + some blog posts from Jared which explains the basic and some advanced morphisms with examples


sellout:
Ah, sorry – can you add a file called “CHANGELOG/revision.md” that says something like “- adds links to Jared Tobin’s blog posts”? _Then_ I can merge.

PR #57, opened 2017-01-18
RecursiveT serializable

rabbitonweb:
Changed Serializable because Quassar connector for Apache Spark needs it to be Serializable
We can alow above since this trait will be removed once we have mutual recursion


sellout:
Hrmm, it looks like this branch isn’t up to date. Needs a merge from master.

PR #60, opened 2017-01-26
Update many dependencies for Matryoshka.

sellout:
Notably,
- crossbuilding on Scalatl 2.11.8 and 2.12.0
- Scalacheck 1.13 (well, pre-1.14, due to an issue not resolved in 1.13), including new `Cogen` and `Shrink` instances
- crossbuilding `scalacheck` and `test` projects on JVM and JS.


tests/shared/src/test/scala/matryoshka/data/fix.scala
```diff
@@ -20,20 +20,18 @@ import matryoshka.exp.Exp
...
 
-class FixSpec extends Specification with CheckAll with AlgebraChecks {
-  "Fix" should {
-    "satisfy relevant laws" in {
-      checkAll(equal.laws[Fix[Exp]])
-    }
-  }
+class FixSpec extends Specification with AlgebraChecks {
+
+  equal.laws[Fix[Exp]].check(Test.Parameters.default)
```
sellout: I’m not very happy with this approach. It seems to work (prints out `+ equal.transitive: OK, passed 100 tests.`), but looks (check the build output) like it’s executed outside of the spec.

jedesah: What's the difference between places where laws are added to the spec like this and ones where they are added like `addFragments(equal.laws[Fix[Exp]])`?

core/shared/src/main/scala/matryoshka/patterns/ListF.scala
```diff
@@ -50,7 +50,7 @@ object ListF {
   def seqIso[A] = AlgebraIso[ListF[A, ?], Seq[A]] {
     case ConsF(h, t) => h +: t
     case NilF()      => Seq.empty
-  } (s => s.headOption.fold[ListF[A, Seq[A]]](NilF())(ConsF(_, s.tail)))
+  } (s => s.headOption.fold[ListF[A, Seq[A]]](NilF())(ConsF(_, s.drop(1))))
```
sellout: Yeah, that’s a good point. I was looking for an `uncons: Option[(A, List[A])]` and was surprised I  couldn’t. Pattern-match it is.

scalacheck/shared/src/main/scala/matryoshka/scalacheck/arbitrary.scala
```diff
@@ -34,7 +34,7 @@ sealed trait ArbitraryInstances0 {
     F(A)
 }
 
-trait ArbitraryInstances extends ArbitraryInstances0 {
+sealed trait ArbitraryInstances extends ArbitraryInstances0 {
```
sellout: Yeah … I actually meant to change this. The problem is that `ArbitraryInstances0` was already `sealed` and `-Xlint` complained about an unsealed extension of a sealed trait. But there is perhaps a better approach … like maybe making `ArbitraryInstances0` `private` instead of `sealed`?

core/shared/src/main/scala/matryoshka/patterns/PotentialFailure.scala
```diff
@@ -34,9 +34,7 @@ final case class PartialFailure[T[_[_]], F[_], E, A] private[patterns](v: F[A])
     extends PotentialFailure[T, F, E, A]
 
 object PotentialFailure {
-  implicit def potentialFailureEqual[T[_[_]], F[_], E: Equal](
-    implicit T: Equal[T[F]], F: Delay[Equal, F]):
-      Delay[Equal, PotentialFailure[T, F, E, ?]] =
+  implicit def potentialFailureEqual[T[_[_]], F[_], E: Equal](implicit T: Equal[T[F]], F: Delay[Equal, F]): Delay[Equal, PotentialFailure[T, F, E, ?]] =
```
sellout: Nope … it’s a requirement for a wart 😕

tests/shared/src/test/scala/matryoshka/data/fix.scala
sellout: Ah, I thought I got rid of all the `addFragments` ones, but looks like I missed a few. The `.check` cases succeed when the `addFragments` ones sometimes give up on too many discarded cases. Also, the `.check` ones give better messages. `addFragments` loses the names of the `Prop`s.

build.sbt
```diff
@@ -201,5 +156,23 @@ lazy val docs = project
...
     micrositeBaseUrl          := "/matryoshka",
     micrositeDocumentationUrl := "/matryoshka/docs/01-Index.html",
-    micrositeHighlightTheme   := "color-brewer"
-  )
+    micrositeHighlightTheme   := "color-brewer")
+
+/** A project just for the console.
+  * Applies only the settings necessary for that purpose.
+  */
+lazy val repl = crossProject dependsOn (core % "test->test;compile->compile") settings standardSettings settings (
```
sellout: I have no idea – some of this (like this part) was taken from @paulp’s previous PR. I assumed it was intended to replace `console <<= console in Test`, which SBT had been spitting out warnings about for a while now.

core/shared/src/main/scala/matryoshka/Hole.scala
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package matryoshka
+
+sealed abstract class Hole
```
sellout: Oh, it was just moved from in “matryoshka/package.scala”, because `-Xlint` complained about classes defined in package objects.

core/shared/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -228,7 +229,7 @@ package object fixedpoint {
     }
 
     // TODO: Would be nice to have this in ApplicativeOps
-    def almostEqual[F[_]: Applicative, A: Equal](a: F[A], b: F[A]): F[Boolean] =
+    def almostEqual[F[_]: Applicative, B: Equal](a: F[B], b: F[B]): F[Boolean] =
```
sellout: It’s only in there because it exists as a helper for `≈`, but yeah, it could be somewhere more general.

core/shared/src/main/scala/matryoshka/patterns/Diffable.scala
```diff
@@ -25,7 +25,7 @@ import scalaz._, Scalaz._
 import simulacrum._
 
 @typeclass trait Diffable[F[_]] { self =>
-  private implicit def self0 = self
+  private implicit def selfʹ: Diffable[F] = self
```
sellout: I forget where exactly, but something in here needs access to a `Diffable` instance, which _can_ be this instance itself, so we make it implicitly available.

core/shared/src/main/scala/matryoshka/patterns/ListF.scala
sellout: Oh, this doesn’t pattern-match because it’s on `Seq`, not `List`. But, this should probably go away anyway. I’ve had a few comments that you just can’t deal with `Seq` in a reasonable way.

sellout:
Updated.

build.sbt
jedesah: Ah, okay.

core/shared/src/main/scala/matryoshka/Hole.scala
jedesah: Ya, still curious what the purpose of the abstract class is, but you're right it's not really relevant to this PR

core/shared/src/main/scala/matryoshka/patterns/Diffable.scala
jedesah: ah, I was searching for references to it but hadn't realized it was implicit. I mean that still sounds weird, but I imagine it makes sense

core/shared/src/main/scala/matryoshka/patterns/PotentialFailure.scala
jedesah: Really...Like a bug or does WartRemover complain about style now?

Or even worse, does Scala style have some subtle effect on semantic? 😨

sellout: Basically, it’s a wart that is currently (has to be?) applied to the source code, and the current regex(?) isn’t smart enough to notice the pattern broken across lines – wartremover/wartremover#226

PR #61, opened 2017-01-30
Improve test output

jedesah:
Integrate law checking with `specs2`


jedesah:
@sellout Let me know what you think of this output.

It's not quite what I would like, but I think it's an improvement. The problem with the previous approach (as I am sure you were aware) is that `scalacheck` is the one responsible for printing the output to the console so it wouldn't ever integrate very well with the rest of the tests being printed out by `specs2`

PR #62, opened 2017-01-30
Upgrade specs2 from `3.8.6` to `3.8.7`

jedesah:
_


jedesah:
oops, @sellout I think I marked you as reviewer instead of assignee on this one. Should be an easy merge.

PR #64, opened 2017-02-16
Rely on sbt-slamdata and slamdata-predef

drostron:
Depends on slamdata/sbt-slamdata#1 and slamdata/slamdata-predef#1

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,118 +1,41 @@
 import de.heikoseeberger.sbtheader.HeaderPlugin
 import de.heikoseeberger.sbtheader.license.Apache2_0
+import org.scalajs.sbtplugin.ScalaJSCrossVersion
 import scoverage._
 import sbt._
 import Keys._
-import org.scalajs.sbtplugin.cross.CrossProject
+import slamdata.CommonDependencies
+import slamdata.SbtSlamData.transferPublishAndTagResources
 
-lazy val checkHeaders = taskKey[Unit]("Fail the build if createHeaders is not up-to-date")
-
-val monocleVersion = "1.4.0"
-val scalazVersion = "7.2.8"
-
-
-lazy val standardSettings = Seq[Setting[_]](
-  headers := Map(
-    "scala" -> Apache2_0("2014–2017", "SlamData Inc."),
-    "java"  -> Apache2_0("2014–2017", "SlamData Inc.")),
-  scalaOrganization := "org.typelevel",
+lazy val standardSettings = commonBuildSettings ++ Seq(
   logBuffered in Compile := false,
   logBuffered in Test := false,
-  outputStrategy := Some(StdoutOutput),
   updateOptions := updateOptions.value.withCachedResolution(true),
-  autoCompilerPlugins := true,
-  autoAPIMappings := true,
   exportJars := true,
   organization := "com.slamdata",
-  resolvers ++= Seq(
-    Resolver.sonatypeRepo("releases"),
-    Resolver.sonatypeRepo("snapshots"),
-    "JBoss repository" at "https://repository.jboss.org/nexus/content/repositories/",
-    "Scalaz Bintray Repo" at "http://dl.bintray.com/scalaz/releases",
-    "bintray/non" at "http://dl.bintray.com/non/maven"),
-  addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.3"),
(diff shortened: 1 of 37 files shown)


build.sbt
```diff
@@ -1,118 +1,47 @@
...
-      email = "contact@slamdata.com",
-      url = new URL("http://slamdata.com"))))
-
-def noPublishSettings = Seq(
-  publish := (),
-  publishLocal := (),
-  publishArtifact := false)
+      "scm:git@github.com:slamdata/matryoshka.git"
+    )
+  ))
```
sellout: What are these fingernail clippings doing here?! 😆

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -16,12 +16,12 @@
...
-import matryoshka.implicits._
-import matryoshka.patterns.EnvT
+import slamdata.Predef._
 
-import scala.{Boolean, Option, PartialFunction}
 import scala.Predef.identity
-import scala.collection.immutable.{List, Nil}
+
+import matryoshka.implicits._
+import matryoshka.patterns.EnvT
```
sellout: We’re in matryoshka, so these imports should be immediately below the Predef.

core/shared/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -201,7 +200,7 @@ package object fixedpoint {
 
     def fromOption[A](opt: Option[A]): Partial[A] = opt.fold(never[A])(now)
 
-    def fromPartialFunction[A, B](pf: scala.PartialFunction[A, B]):
+    def fromPartialFunction[A, B](pf: PartialFunction[A, B]):
```
sellout: Hrmm, wondering if we want `PartialFunction` in our Predef … (not that it’d be changed in this PR)

project/plugins.sbt
```diff
@@ -1,15 +1,7 @@
...
-addSbtPlugin("com.dwijnand"          % "sbt-travisci"    % "1.0.0")
-addSbtPlugin("com.eed3si9n"          % "sbt-unidoc"      % "0.3.3")
-addSbtPlugin("org.wartremover"       % "sbt-wartremover" % "1.2.1")
+addSbtPlugin("io.get-coursier"       % "sbt-coursier"   % "1.0.0-M15")
+addSbtPlugin("com.fortysevendeg"     % "sbt-microsites" % "0.3.3")
+addSbtPlugin("org.scala-js"          % "sbt-scalajs"    % "0.6.14")
+addSbtPlugin("org.scoverage"         % "sbt-scoverage"  % "1.5.0")
+addSbtPlugin("com.typesafe.sbt"      % "sbt-site"       % "1.1.0")
+addSbtPlugin("com.slamdata"          % "sbt-slamdata"   % "0.0.1")
+addSbtPlugin("com.eed3si9n"          % "sbt-unidoc"     % "0.3.3")
```
sellout: Should any of these _not_ be in sbt-slamdata?

Maybe the tut ones … but even then I’m not sure. It should be minimal effort to tut any of our code.

Again, not for this PR, but for a subsequent sbt-slamdata update (or you could cascade it all through this series of PRs).

build.sbt
drostron: 😆, _sweep_ _sweep_

project/plugins.sbt
drostron: They might all eventually belong in sbt-slamdata. I was thinking that we'll have a set of default plugins that are brought in automatically and another set offered in a similar fashion as dependencies so they can selected per project.

core/shared/src/main/scala/matryoshka/instances/fixedpoint/package.scala
drostron: 🤔

drostron:
@sellout Ready for another look

project/plugins.sbt
sellout: BTW, yesterday @rabbitonweb mentioned having tut docs for QScript, etc. which made me _more_ inclined to get these into sbt-slamdata. However, since something in like sbt-site or somewhere doesn’t work with git worktree, I don’t want it in there until that’s fixed – not being able to use git worktree easily in Matryoshka is already a problem, and I don’t want it to spread.

drostron: Ok, good to be aware of that issue. If it does get moved to sbt-slamdata we'll at least go with an opt in approach until addressed.

PR #65, opened 2017-02-16
Fix typo in documentation

jedesah:


core/shared/src/main/scala/matryoshka/Based.scala
```diff
@@ -21,7 +21,7 @@ import simulacrum.typeclass
 /** Provides a type describing the pattern functor of some {co}recursive type
   * `T`. For standard fixed-point types like [[matryoshka.data.Fix]],
   * `Patterned[Fix[F]]#Base` is simply `F`. However, directly recursive types
-  * generally have a less obivous pattern functor. E.g., `Patterned[Cofree[F,
```
sellout: So, also in this comment, could you also replace `Patterned` with `Based`? 😄

jedesah: done!

PR #67, opened 2017-02-24
Make repeatedly tail-recursive

edmundnoble:
Now's the second time I've told someone to steal tail-recursive repeatedly from slate. Fixing at the source would be nicer.

### DIFF
--- a/core/shared/src/main/scala/matryoshka/package.scala
+++ b/core/shared/src/main/scala/matryoshka/package.scala
@@ -641,7 +641,7 @@ package object matryoshka {
     *
     * @group algtrans
     */
-  @annotation.tailrec
+  @tailrec
   final def repeatedly[A](f: A => Option[A])(expr: A): A =
     f(expr) match {
       case None => expr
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -641,8 +641,12 @@ package object matryoshka {
     *
     * @group algtrans
     */
-  def repeatedly[A](f: A => Option[A]): A => A =
-    expr => f(expr).fold(expr)(repeatedly(f))
+  @annotation.tailrec
```
sellout: We use a custom Predef, which provides `tailrec` (but doesn’t include `_root_.scala`), so you don’t need to qualify it.

wemrysi:
@sellout Any idea why travis didn't build this? `spec.scala` fails to build on [L104](https://github.com/slamdata/matryoshka/blob/master/tests/shared/src/test/scala/matryoshka/spec.scala#L104)

sellout:
I merged because it was all green. *sigh*

PR #68, opened 2017-02-26
Add monadic zygomorphism

kanterov:
I'm developing compiler with matryoshka, my original motivation is to annotate tree with types, and then compile, both phases operate under `M = Error \/ ?`. Basically, I want to do zygomorphism where both algebras are effectful.

I'm very new to recursion schemes, but I have a feeling there is something wrong because I wasn't able to describe `zygoM` in terms of `gcataM`. I'm not sure that this function is called `zygoM`, because probably it should be:

```scala
  def zygoM[A, B, M[_]: Monad]
    (t: T)
    (f: Algebra[Base, B], g: GAlgebraM[(B, ?), M, Base, A])
    (implicit BF: Traverse[Base])
      : M[A]
```

But then I'm not sure what is the right function for the particular use case I had.

I didn't add any tests because I'm not yet sure this is the right thing. I would be glad to get any help or feedback on this.

### DIFF
--- a/core/shared/src/main/scala/matryoshka/Recursive.scala
+++ b/core/shared/src/main/scala/matryoshka/Recursive.scala
@@ -50,13 +50,14 @@ trait Recursive[T] extends Based[T] {
     g(loop(t).copoint)
   }
 
-  def gcataM[W[_]: Comonad: Traverse, M[_]: Monad, A](
-    t: T)(
-    k: DistributiveLaw[Base, W], g: GAlgebraM[W, M, Base, A])(
-    implicit BT: Traverse[Base]):
-      M[A] = {
+  def gcataM[W[_]: Comonad : Traverse, M[_]: Monad, A]
+    (t: T)
+    (w: DistributiveLaw[Base, (M ∘ W)#λ], g: GAlgebraM[W, M, Base, A])
+    (implicit BT: Traverse[Base])
+      : M[A] = {
     def loop(t: T): M[W[Base[W[A]]]] =
-      project(t).traverse(loop(_) >>= (_.traverse(g) ∘ (_.cojoin))) ∘ (k(_))
+      project(t).traverse(loop(_) >>=
+        (_.traverse(g) ∘ (_.cojoin))) ∘ (_ ∘ (_.point[M])) >>= (w(_))
 
     loop(t) ∘ (_.copoint) >>= g
   }
@@ -471,10 +472,10 @@ object Recursive {
         : A =
       typeClassInstance.gcata[W, A](self)(k, g)
     def gcataM[W[_]: Comonad: Traverse, M[_]: Monad, A]
-      (k: DistributiveLaw[F, W], g: GAlgebraM[W, M, F, A])
+      (w: DistributiveLaw[F, (M ∘ W)#λ], g: GAlgebraM[W, M, F, A])
       (implicit BT: Traverse[F])
         : M[A] =
-      typeClassInstance.gcataM[W, M, A](self)(k, g)
+      typeClassInstance.gcataM[W, M, A](self)(w, g)
     def elgotCata[W[_]: Comonad, A]
       (k: DistributiveLaw[F, W], g: ElgotAlgebra[W, F, A])
       (implicit BF: Functor[F])
--- a/core/shared/src/main/scala/matryoshka/Recursive.scala
+++ b/core/shared/src/main/scala/matryoshka/Recursive.scala
@@ -73,6 +73,20 @@ trait Recursive[T] extends Based[T] {
     g(loop(t))
   }
 
+  def elgotCataM[W[_]: Comonad : Traverse, M[_]: Monad, A]
+    (t: T)
+    (k: DistributiveLaw[Base, (M ∘ W)#λ], g: ElgotAlgebraM[W, M, Base, A])
+    (implicit BT: Traverse[Base])
+      : M[A] = {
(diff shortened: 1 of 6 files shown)


kanterov:
Just realized that I my particular case I need something like

```scala
  def elgotZygoM[A, B, F[_]: Traverse, M[_]: Monad]
    (t: Fix[F])
    (f: AlgebraM[M, F, B], g: ElgotAlgebraM[(B, ?), M, F, A])
      : M[A]
```

sellout:
@kanterov Feel free to add that, too 😄

kanterov:
I added `zygoElgotM`, didn't touch tests yet, but it typechecked :)

kanterov:
@sellout I added tests. I had to extract zygomorphism tests because compilation time of `specs.scala` is terrible. Can you please take a look?

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -101,13 +115,37 @@ trait Recursive[T] extends Based[T] {
...
+  def zygoM[A, B, M[_]: Monad]
+    (t: T)
+    (f: AlgebraM[M, Base, B], g: GAlgebraM[(B, ?), M, Base, A])
+    (implicit BT: Traverse[Base])
+      : M[A] = {
+    val k = distZygoM(f, distApplicative[Base, M])
+
+    def loop[W[_]: Comonad : Traverse]
+      (t: T)
+      (g: GAlgebraM[W, M, Base, A], k: DistributiveLaw[Base, (M ∘ W)#λ])
```
sellout: You can get rid of these parameters, because `loop` always uses the same `g` and `k`.

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -101,13 +115,37 @@ trait Recursive[T] extends Based[T] {
...
     (implicit BF: Functor[Base])
       : A =
     elgotCata[(B, ?), A](t)(distZygo(f), g)
 
+  def elgotZygoM[A, B, M[_]: Monad]
+    (t: T)
+    (f: AlgebraM[M, Base, B], g: ElgotAlgebraM[(B, ?), M, Base, A])
+    (implicit BT: Traverse[Base])
+      : M[A] =
+    elgotCataM[A, B, (B, ?), M](t)(distZygoM(f, distApplicative[Base, M]), g)
```
sellout: Yeah, it’s interesting that you can use `distZygoM` here in `elgotCataM`, but not in `gCataM` above.

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -402,6 +402,18 @@ package object matryoshka {
     *
     * @group dist
     */
+  def distZygoM[F[_]: Functor, M[_]: Monad, B](
```
sellout: Nice.

kanterov:
@sellout I've updated `zygoM`, I had problem with implicit resolution, so I had to introduce `W[_]` before, I've cleaned it up and extract signature of `gcataM` as it probably should be. Could you please take a look?

kanterov:
Signing CLA is WIP, need to check with my employer first, but it shouldn't be a problem.

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -101,13 +115,42 @@ trait Recursive[T] extends Based[T] {
...
     gcata[(B, ?), A](t)(distZygo(f), g)
 
+  def zygoM[A, B, M[_]: Monad]
+    (t: T)
+    (f: AlgebraM[M, Base, B], g: GAlgebraM[(B, ?), M, Base, A])
+    (implicit BT: Traverse[Base])
+      : M[A] = {
+
+    // NB: Probably this is how `gcataM` should look like
+    def gcataM[W[_]: Comonad : Traverse]
```
sellout: I think this looks reasonable. How about you replace the existing `gcataM` and see if everything still works with this new def?

kanterov:
@sellout squashed, signed CLA, generalized `gcataM`, can you please take a look again? :)

sellout:
Trying to trigger a Travis build …

sellout:
I know this _looks_ mergable, but our merge script says it’s not. Can you merge master into this branch? Then we should be good.

kanterov:
@sellout I've rebased latest master on this branch.

Are these scripts public? Your release process looks awesome. 

By the way, I'm not familiar with it, but probably this change needs to be documented as breaking.

sellout:
Breaking because of the change to `gcataM`? This is what I get for not having better test coverage … I had _assumed_ that there were already uses of `gcataM` in definitions of other monadic, but alas that’s not true.

So, the gist of the breaking change is that `def cataM(t: T)(f: AlgebraM) = gcataM(t)(distCata, f)` will no longer work because we need to do something about the new `M ∘` in the `gcataM` `DistributiveLaw`, right?

sellout:
Also, our merge script is at https://github.com/slamdata/devtools/blob/master/bin/merge, and any of the publishing releases stuff is in https://github.com/slamdata/sbt-slamdata it’s not perfect, but I’m putting together a blog post that describes it and what parts of it we do & don’t like currently.

kanterov:
Yes, true, but I didn't find any open source codebase using `gcataM`, and it's possible to write a function from `DistributiveLaw[F, W]` to `DistributiveLaw[F, M ∘ W]` for any `M[_]: Applicative`, so I see it mostly as a change breaking binary compatibility.

PR #72, opened 2017-03-08
Define more operations in terms of cata, ana, and hylo.

sellout:
This is especially useful for Mu/Nu, since these operations will now use the specialized ana/cata rather than the naïve implementations.

Also greatly reduces the number of places we use general recursion.

### DIFF
--- a/core/shared/src/main/scala/matryoshka/Corecursive.scala
+++ b/core/shared/src/main/scala/matryoshka/Corecursive.scala
@@ -32,7 +32,7 @@ trait Corecursive[T] extends Based[T] {
     (f: CoalgebraM[M, Base, A])
     (implicit BT: Traverse[Base])
       : M[T] =
-    f(a).flatMap(_.traverse(anaM(_)(f))) ∘ (embed(_))
+    hyloM[M, Base, A, T](a)(embed(_).point[M], f)
 
   def gana[N[_]: Monad, A]
     (a: A)
@@ -45,12 +45,8 @@ trait Corecursive[T] extends Based[T] {
     a: A)(
     k: DistributiveLaw[N, Base], f: GCoalgebraM[N, M, Base, A])(
     implicit BT: Traverse[Base]):
-      M[T] = {
-    def loop(x: N[Base[N[A]]]): M[T] =
-      k(x).traverse(_.join.traverse(f) >>= loop) ∘ (embed(_))
-
-    f(a) ∘ (_.point[N]) >>= loop
-  }
+      M[T] =
+    ghyloM[Id, N, M, Base, A, T](a)(distCata, k, embed(_).point[M], f)
 
   def elgotAna[N[_]: Monad, A](
     a: A)(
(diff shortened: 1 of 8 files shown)


sellout:
@paulp I’m guessing this should remove ~19 of the 68 warnings from the Recursion wart. It’s a start.

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -32,35 +32,27 @@ trait Recursive[T] extends Based[T] {
...
-    g(loop(t).copoint)
-  }
+      : A =
+    cata[W[A]](t)(fwa => k(fwa.map(_.cojoin)).map(g)).copoint
 
-  def gcataM[W[_]: Comonad : Traverse, M[_]: Monad, A]
+  def gcataM[W[_]: Comonad: Traverse, M[_]: Monad, A]
     (t: T)
-    (w: DistributiveLaw[Base, (M ∘ W)#λ], g: GAlgebraM[W, M, Base, A])
+    (w: DistributiveLaw[Base, W], g: GAlgebraM[W, M, Base, A])
```
paulp: Heh.

core/shared/src/main/scala/matryoshka/Corecursive.scala
```diff
@@ -25,7 +25,7 @@ trait Corecursive[T] extends Based[T] {
   def embed(t: Base[T])(implicit BF: Functor[Base]): T
 
   def ana[A](a: A)(f: Coalgebra[Base, A])(implicit BF: Functor[Base]): T =
-    embed(f(a) ∘ (ana(_)(f)))
+    hylo(a)(embed, f)
```
paulp: Such a big improvement this way.

paulp:
Very much an improvement. From my POV the entire endeavor is much clearer when the hylomorphism takes center stage and the rest is understood relative to that (e.g. ana and cata as degenerate hylomorphisms) where possible. So to see the implementations take that approach is excellent.

Baccata:
Wow this makes so much sense ! Really helps wrapping my head around the concepts. Seems like it will make for a much simpler expression of why recursion schemes matter.

sellout:
Yeah, I often explain things like `cata(f) === hylo(f, project)`, and the implementations are actually identical, so I don‘t know why I didn’t actually define it this way before. The Recursion wart really helps highlight the things we want to avoid, though 😆

sellout:
@paulp BTW, there are at least a handful of other cases I’d like to fix in this PR if you don’t mind waiting a bit. The shorter warning list made a few other changes obvious.

tscholak:
just out of curiosity, for `cata(f) === hylo(f, project)`, are there impacts on speed, either during compile or run time?

sellout:
@tscholak I haven’t benchmarked it, the implementation is pretty much identical. The only difference is that `project` is passed as a parameter to each recursive call and I’m not sure what the overhead of that might be. Also, there is the indirection of one more function call – cata -> hylo.

sellout:
@paulp Ok, I think that’s all for the time being.

sellout:
@paulp Ok, merged in master and made a couple more small changes. Should be ready for merge for reals.

paulp:
It would be great to leverage this very library to compare the precise operation counts of various approaches. No reason we should have to speculate about what's happening under the hood.

PR #78, opened 2017-04-11
Repeatedly func

rabbitonweb:
`repeatedly` returns one more time a function value (instead of being a method). 

also added `ginterpret`

### DIFF
--- a/core/shared/src/main/scala/matryoshka/package.scala
+++ b/core/shared/src/main/scala/matryoshka/package.scala
@@ -681,8 +681,8 @@ package object matryoshka {
     *
     * @group algtrans
     */
-  @tailrec
-  final def repeatedly[A](f: A => Option[A])(expr: A): A =
+  final def repeatedly[A]: (A => Option[A]) => A => A =
+    (f: A => Option[A]) => (expr: A) =>
     f(expr) match {
       case None => expr
       case Some(e) => repeatedly(f)(e)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -681,8 +681,8 @@ package object matryoshka {
     *
     * @group algtrans
     */
-  @tailrec
```
sellout: I would like this to be a function, but the problem is that we need this to be tail recursive.

We could actually rewrite this as a coalgebra, like:
```scala
final def repeatedlyƒ[A](f: A => Option[A]): Coalgebra[A \/ ?, A] = f(expr) \/> expr

final def repeatedly[A](f: A => Option[A]): A => A =
  _.ana[Nu](repeatedlyƒ(f)).unsafePerformSync
```
We have some stack-safety tests of `Partial` that make me think this works, although no performance promises.

But then maybe we can rewrite stuff to use `repeatedlyƒ` instead of `repeatedly`, and track partiality through more of the code.

@edmundnoble – thoughts?

core/shared/src/main/scala/matryoshka/patterns/package.scala
```diff
@@ -54,6 +54,9 @@ package object patterns {
       AlgebraM[M, CoEnv[A, F, ?], B] =
     ginterpretM[Id, M, F, A, B](f, φ)
 
+  def ginterpret[W[_], F[_], A, B](f: A => Id[B], φ: GAlgebra[W, F, B]): GAlgebra[W, CoEnv[A, F, ?], B] =
```
sellout: 👍🏾

sellout:
So, Matryoshka has a new definition of `repeatedly` now that basically requires it to return `Partial[A]`, tracking the general recursion explicitly, so if this PR (including the description) could be updated to _only_ have `ginterpret` in it, that’d be great, and we can get it merged and gone.

rabbitonweb:
:+1: but Monday after spark summit

sellout:
@rabbitonweb Ping.

PR #79, opened 2017-04-15
Fix the Birecursive type class.

### LINKED ISSUE #44: Add `Birecursive` type class

sellout:
Fixes #44.

Replaces all instances and constraints that were both `Recursive` and
`Corecursive` with `Birecursive`. Move type class operations with both
constraints to the `Birecursive` type class (including overloading some
operations that have better implementations with both constraints).

Also added `gcataZygo`, which enables an overload of `paraZygo` in
`Birecursive`. And added `Delay.apply`.


core/shared/src/main/scala/matryoshka/Birecursive.scala
```diff
@@ -16,18 +16,160 @@
...
-  *     #44 is resolved. Define an instance of [[BirecursiveT]] if possible,
-  *     otherwise define separate [[Recursive]] and [[Corecursive]] instances.
   */
-// TODO: Once this type class is actually usable, operations in `Recursive` and
-//       `Corecursive` that also have a `Corecursive.Base[T, Base]` constraint
-//       should be moved here.
-trait Birecursive[T] extends Recursive[T] with Corecursive[T]
+trait Birecursive[T] extends Recursive[T] with Corecursive[T] { self =>
+  implicit val rec: Recursive.Aux[T, Base] = self
+  implicit val corec: Corecursive.Aux[T, Base] = self
```
sellout: These implicits allow the constraints used in definition here to be satisfied.

core/shared/src/main/scala/matryoshka/implicits/package.scala
```diff
@@ -23,15 +23,28 @@ package object implicits
     with Merge.ToMergeOps
     with Recursive.ToRecursiveOps {
 
+  implicit def toRecursive[T, F[_]](implicit T: Birecursive.Aux[T, F])
+      : Recursive.Aux[T, F] =
+    T.rec
+
+  implicit def toCorecursive[T, F[_]](implicit T: Birecursive.Aux[T, F])
+      : Corecursive.Aux[T, F] =
+    T.corec
```
sellout: And these implicits expose the ones defined in the type class, so now everything can resolve Rec/Corec operations given a Birec constraint.

sellout:
In addition to the people I added as reviewers, I figure @edmundnoble and @milessabin may also have thoughts about the way I managed to get `<super>.Aux` to resolve from `<sub>.Aux`.

core/shared/src/main/scala/matryoshka/Birecursive.scala
milessabin: Given the later `toRecursive` and `toCorecursive` definitions, do these two actually need to be implicit?

core/shared/src/main/scala/matryoshka/Recursive.scala
```diff
@@ -165,6 +165,16 @@ trait Recursive[T] extends Based[T] {
       A =
     gcata[Cofree[H, ?], A](t)(distGHisto(g), f)
 
+  def gcataZygo[W[_]: Comonad, A, B]
```
drostron: Is there a corresponding test?

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -177,34 +176,32 @@ package object matryoshka {
   }
 
   // TODO: Move this to `Birecursive` once that works.
-  def birecursiveIso[T, F[_]: Functor]
-    (implicit TR: Recursive.Aux[T, F], TC: Corecursive.Aux[T, F]) =
+  def birecursiveIso[T, F[_]: Functor](implicit T: Birecursive.Aux[T, F]) =
```
sellout: Umm, nothing, I think 😄 I should do that.

core/shared/src/main/scala/matryoshka/Recursive.scala
sellout: Indirectly, because of https://github.com/slamdata/matryoshka/pull/79/files/2641f7f624f4e15d3df1974b4532f874aac94021#diff-08bebdfb63dd73c8d3dab8adcf5d9777R73

PR #82, opened 2017-04-23
Greatly reduce the general recursion used.

### LINKED ISSUE #24: Add "How to install instructions"

### LINKED ISSUE #59: README should mention the need to import fixpoint types
To use `Fix` etc., it is insufficient to import only

```
import matryoshka._
import matryoshka.implicits._
```

An explicit import of `matryoshka.data.Fix` etc. is required. It would be helpful if the README said so more prominently; right now, this is buried in the `eval` example.

sellout:
When the next version of WartRemover is released, this will be enforced, but it
can be checked with a snapshot version for the time being. This reduces 40+
instances of recursion to now only 6 (labeled) instances. And this is after a
refactoring of the library already eliminated at least 20 other instances.

Transforms of unfolds (`transAna`, etc.) have moved to `Corecursive`, because
that's the only way to make it possible for `Birecursive` to override them
properly. This required a new `implicit class RecursiveOps` to hold the
enrichments of those unfolds.

Generalized a bunch of things that previously specified types like `Nat` to use
`[N](implicit N: Recursive.Aux[N, Option])`.

Renamed `NelF` to `AndMaybe`. And added fixed-point representations of
NonEmptyList.

`repeatedly`, which truly uses general recursion now correctly labels it with
the `Partial` monad.

Fixes #24, fixes #59, which improve the README.


sellout:
There are some other changes I would like to make here – e.g., moving various implicits from `matryoshka.instances.fixedpoint` to `matryoshka.patterns` and some more generalization, but I figured the diff was big enough as it is.

core/shared/src/main/scala/matryoshka/instances/fixedpoint/package.scala
```diff
@@ -205,28 +281,38 @@ package object fixedpoint {
...
+      }.map(_.step).merge
 
     /** Run to completion (if it completes).
       */
-    // TODO: Alternatively `self.cata[A](_.merge)`, but that's not tailrec.
-    @tailrec final def unsafePerformSync: A = self.project match {
-      case -\/(a) => a
-      case \/-(p) => p.unsafePerformSync
-    }
+    final def unsafePerformSync: A = self.cata[A](_.merge)
```
wemrysi: Is `tailrec` being achieved via some other mechanism here? If not, it would seem to reduce the utility of this type, no?

tests/shared/src/test/scala/matryoshka/instances/fixedpoint/partial.scala
```diff
@@ -50,46 +48,43 @@ class PartialSpec extends Specification with ScalazMatchers with ScalaCheck {
...
-        beLeftDisjunction(4)
+    "return exactly at the end" >> prop { (n: Conat, i: Int) =>
+      n.transAna[Partial[Int]](Partial.delay(i)).runFor(n) must
+        beLeftDisjunction(i)
     }
   }
 
   "unsafePerformSync" should {
-    "return now immediately" in {
+    "return novw immediately" in {
```
wemrysi: Typo?

core/shared/src/main/scala/matryoshka/data/option.scala
```diff
@@ -25,6 +25,27 @@ trait OptionInstances {
   implicit def optionBirecursive[A]
       : Birecursive.Aux[Option[A], Const[Option[A], ?]] =
     id.idBirecursive[Option[A]]
+
+  implicit val optionDelayEqual: Delay[Equal, Option] =
+    Delay.fromNT(λ[Equal ~> (Equal ∘ Option)#λ](eq =>
+      Equal.equal {
```
sellout: I’ll try this, but I _think_ it runs into the same infinite regression thing when trying to resolve, say,  `Mu[Option]`.

core/shared/src/main/scala/matryoshka/instances/fixedpoint/package.scala
sellout: Yeah – it’s being handled by the fixed-point. So, for `Cofree[A \/ ?, B]`, that’d be `Trampoline`, for `Nu`, it’d be the delayed call, etc. For `Mu` and `Fix`, well, you’re going to have trouble constructing a very partial `Mu[A \/ ?]` before you ever get around to running it 😆

But I’m not certain about this, so I should probably add a nice test or two (with a comment that it depends on your stack size).

core/shared/src/main/scala/matryoshka/data/option.scala
```diff
@@ -25,6 +25,27 @@ trait OptionInstances {
...
+
+  implicit val optionDelayEqual: Delay[Equal, Option] =
+    new Delay[Equal, Option] {
+      def apply[A](a: Equal[A]) = {
+        implicit val aʹ: Equal[A] = a
+        Equal[Option[A]]
+      }
+    }
+
+  implicit val optionDelayOrder: Delay[Order, Option] =
```
wemrysi: Apologies for not being more explicit, but my comment on `Equal[Option[A]]` also applies to `Order[Option[A]]` and `Show[Option[A]]`.

sellout: Of course it does – I just had too much tunnel vision to see the other code.

sellout:
Updated again.

PR #90, opened 2017-07-10
Generalize the various attribute* algebras.

wemrysi:
Generalizes the various `attribute*` algebras to work for any `[T, F[_], A] => Birecursive.Aux[T, EnvT[A, F, ?]]`.

### DIFF
--- a/core/shared/src/main/scala/matryoshka/package.scala
+++ b/core/shared/src/main/scala/matryoshka/package.scala
@@ -108,13 +108,13 @@ package object matryoshka {
     * in bottom-up fashion.
     * @group algebras
     */
-  type ElgotAlgebraicTransformM[W[_], M[_], T, F[_], G[_]] = W[F[T]] => M[G[T]]
+  type AlgebraicElgotTransformM[W[_], M[_], T, F[_], G[_]] = W[F[T]] => M[G[T]]
 
   /** Transform a structure `F` contained in `W`, to a structure `G`,
     * in bottom-up fashion.
     * @group algebras
     */
-  type ElgotAlgebraicTransform[W[_], T, F[_], G[_]] = W[F[T]] => G[T]
+  type AlgebraicElgotTransform[W[_], T, F[_], G[_]] = W[F[T]] => G[T]
 
   /** Transform a structure `F` to a structure `G` containing values in `N`,
     * in top-down fashion, accumulating effects in the monad `M`.
@@ -665,7 +665,7 @@ package object matryoshka {
       def apply[F[_]: Functor, A]
         (f: ElgotAlgebraM[W, M, F, A])
         (implicit W: Comonad[W], M: Functor[M], T: Recursive.Aux[T, EnvT[A, F, ?]])
-          : ElgotAlgebraicTransformM[W, M, T, F, EnvT[A, F, ?]] =
+          : AlgebraicElgotTransformM[W, M, T, F, EnvT[A, F, ?]] =
         node => f(node ∘ (_ ∘ (_.project.ask))) ∘ (a => EnvT((a, node.copoint)))
     }
   }
@@ -681,7 +681,7 @@ package object matryoshka {
       def apply[F[_]: Functor, A]
         (f: ElgotAlgebra[W, F, A])
         (implicit W: Comonad[W], T: Recursive.Aux[T, EnvT[A, F, ?]])
-          : ElgotAlgebraicTransform[W, T, F, EnvT[A, F, ?]] =
+          : AlgebraicElgotTransform[W, T, F, EnvT[A, F, ?]] =
         attributeElgotM[W, Id, T](f)
     }
   }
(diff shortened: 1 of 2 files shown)


sellout:
This is great – is there a use case in Quasar, or is it just a deficiency you noticed?

wemrysi:
@sellout Yeah, `quasar.sst.StructuralType` is a newtyped `Cofree` that I'd use these generalized versions on if they existed.

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -561,16 +561,33 @@ package object matryoshka {
...
-    fa => f(fa ∘ (_.head)) ∘ (Cofree(_, fa))
+  object attributeAlgebraM {
+    def apply[T] = new PartiallyApplied[T]
+
+    final class PartiallyApplied[T] {
+      def apply[F[_]: Functor, M[_]: Functor, A]
+        (f: AlgebraM[M, F, A])
+        (implicit T: Birecursive.Aux[T, EnvT[A, F, ?]])
+          : AlgebraM[M, F, T] =
+        fa => f(fa ∘ (_.project.ask)) ∘ (a => EnvT((a, fa)).embed)
```
sellout: I haven’t thought through if this is actually possible, but could this be
```scala
(implicit T: Recursive.Aux[T, EnvT[A, F, ?]])
  : TransformM[M, F, EnvT[A, F, ?], T]
```
and just remove the `.embed`? I’m not sure it really gains us anything (`transCataM` will need the `Corecursive` instance anyway), but I try to keep the constraints down.

tests/shared/src/test/scala/matryoshka/spec.scala
```diff
@@ -1018,27 +1018,29 @@ class MatryoshkaSpecs extends Specification with ScalaCheck with ScalazMatchers
   }
 
   "Attr" >> {
+    type T[A] = Cofree[Exp, A]
+
     "attrSelf" >> {
       "annotate all" >> Prop.forAll(expGen) { exp =>
-        exp.cata(attrSelf[Mu[Exp], Exp]).elgotPara(universe) must
-          equal(exp.elgotPara(universe).map(_.cata(attrSelf[Mu[Exp], Exp])))
+        exp.cata(attrSelf[T[Mu[Exp]]].apply).elgotPara(universe) must
```
sellout: Meh. I think it’s unavoidable until we have scala/scala#5108, but I really hate this `.apply` stuff.

core/shared/src/main/scala/matryoshka/package.scala
sellout: (similarly for the other things that are implemented in terms of this)

tests/shared/src/test/scala/matryoshka/spec.scala
wemrysi: Agreed, I considered renaming the inner `apply` to something like `f`, but ultimately decided against it for consistency with other usages of this technique.

sellout: Yeah, I think it “has to be” `apply`, simply for the cases where we do apply them directly and want it to look like a function.

wemrysi: Aye, but the inner `apply` in `attrSelf` takes no arguments, so if you ever want to invoke it directly you'll have to use `apply` or completely fill in both type parameter lists 😞. Anyhow, I think the consistency is worth, so /shrug.

core/shared/src/main/scala/matryoshka/package.scala
wemrysi: It looks like it should be possible, I'll give it a go. And 👍 for keeping constraints down.

wemrysi: @sellout Does the `Algebra` in `attributeAlgebraM` refer to the argument or the result? If the latter, I propose renaming to just `attributeM` if the transform approach works out (or `attributeTransM` if we must).

wemrysi: Ah, given `attributeElgotM` I'm guessing it refers to the argument, so nevermind.

core/shared/src/main/scala/matryoshka/package.scala
```diff
@@ -104,6 +104,18 @@ package object matryoshka {
     */
   type AlgebraicGTransform[W[_], T, F[_], G[_]]        = F[W[T]] => G[T]
 
+  /** Transform a structure `F` contained in `W`, to a structure `G`,
+    * in bottom-up fashion.
+    * @group algebras
+    */
+  type ElgotAlgebraicTransformM[W[_], M[_], T, F[_], G[_]] = W[F[T]] => M[G[T]]
```
sellout: I would name these `AlgebraicElgotTransform`[`M`], to match `CoalgebraicElgotTransform`, `AlgebraicGTransform`, etc.

wemrysi: Of course ;)

PR #92, opened 2017-09-20
Add automatic differentiation example

zainab-ali:
This is a port of Jared Tobin's [Automatic Differentiation](https://jtobin.io/ad-via-recursion-schemes)

### DIFF
--- a/tests/shared/src/test/scala/matryoshka/example/mathExpr.scala
+++ b/tests/shared/src/test/scala/matryoshka/example/mathExpr.scala
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2014–2017 SlamData Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


tests/shared/src/test/scala/matryoshka/example/mathExpr.scala
```diff
@@ -0,0 +1,126 @@
...
+    case OneF() => zero
+    case NegateF((_, a)) => neg(a)
+    case SumF((_, l), (_, r)) => add(l, r)
+    case ProductF((l, ll), (r, rr)) => add(prod(l, rr),  prod(r, ll))
+    case ExpF((x, xx)) => prod(e(x), xx)
+  }
+
+  def diff(expr: Expr): Expr = expr.para(diffGAlgebra)
+
+  def adAlgebra(x: Double): Algebra[MathExprF, (Double, Double)] = {
```
wemrysi: If you change this to a `GAlgebra[(Double, ?), MathExprF, Double]` you could compute the derivative via a zygomorphism in `automaticDifferentiation` using `evalAlgebra` and avoid having to duplicate it here.

zainab-ali: Thanks!  I was thinking along the same lines, and couldn't quite work out why Jared Tobin's examples didn't do this.

Fristi:
Awesome stuff! :)

PR #97, opened 2018-05-08
sbt 0.13.17

xuwei-k:


djspiewak:
We're on 1.0 now, thanks to @davidhoyt, and also thanks to the fact that I've completely failed at keeping up with this PR queue.

